PR #1, opened 2016-12-22
Initial template thoughts

klumba12:


docs/template.md
```diff
@@ -1,7 +1,25 @@
-# qgrid 
+# qgrid template
 ## markup
 ```html
+<q-grid rows="data" >
```
klumba12: what name we want to have q-grid, qgrid, q-table, qtable or smth different?

docs/template.md
```diff
@@ -1,7 +1,25 @@
...
+        <q-grid:head> 
+            <q-grid:td> 
+                <i title="{{$column.name}}">{{$column.label}}</i>
+            </q-grid:td>
+        </q-grid:head>
+        <q-grid:body>
+            <q-grid:tr>
```
klumba12, line 11: are we going to handle row templating? how it should looks like?

docs/template.md
```diff
@@ -1,7 +1,25 @@
-# qgrid 
+# qgrid template
 ## markup
 ```html
+<q-grid rows="data" >
+        <q-grid:head>
```
klumba12, line 5: if there won't be row templating are we need head body? or just can use q-grid:th, q-grid:td?

docs/template.md
```diff
@@ -1,7 +1,25 @@
...
+                <i title="{{$column.name}}">{{$column.label}}</i>
+            </q-grid:td>
+        </q-grid:head>
+        <q-grid:body>
+            <q-grid:tr>
+            </q-grid:tr>
+            <q-grid:td key="firstName">
+                <b>{{$value}}</b>
+            </q-grid:td>
+            <q-grid:td key="dateBirth">
```
klumba12, line 16: What will be look of editable template? Something like triggers in wpf?
My variants:
*<q-grid:td-in-editmode>
*<q-grid:td:edit>
*<q-grid:td on='edit'>
*<q-grid:td when='$mode === "edit"'>
*<q-grid:td on:edit='true'>
*<q-grid:td on:mode='edit'>   ?

docs/template.md
```diff
@@ -1,7 +1,25 @@
-# qgrid 
+# qgrid template
 ## markup
 ```html
+<q-grid rows="data" >
+        <q-grid:head> 
+            <q-grid:td> 
+                <i title="{{$column.name}}">{{$column.label}}</i>
```
klumba12, line 7: is $column good name enought? how we can propogate properties with other way?

docs/template.md
```diff
@@ -1,7 +1,25 @@
-# qgrid 
+# qgrid template
 ## markup
 ```html
+<q-grid rows="data" >
+        <q-grid:head> 
+            <q-grid:td>
```
klumba12, line 6: what if we want to have multi-column layout how to do it?

docs/template.md
vkorolev: I like something like this option:
`<q-grid:td when='$mode === "edit"'>`
It's more flexible and its semantics allows to use expressions, which can be easily combined like:
`<q-grid:td when="$column.isSelected && $row.name == 'John'">`

klumba12: What will be perfomance impact?

docs/template.md
klumba12: It can be done through js columns property easily dont know if we need it in templates?

docs/template.md
klumba12: And i defently like $row property

docs/template.md
titovmx: I think, more cases are covered by cell templating. Can you suggest row templating case?

docs/template.md
vkorolev: Of course, if it's a watcher, it will affect performance badly. Probably, we can have only restricted set of events like 'onRowSelect', 'onModeChanged' to reduce load.

docs/template.md
klumba12: e.g. highlight specific row

vkorolev: If we're going to use it as grid, the only thing, i can imagine is using UL and LI for rows.
I'm not sure, it's a good or even possible case

vkorolev: I think, highlighting may be resolved by CSS classes.

docs/template.md
klumba12: thats why i like on:mode='edit' e.g. need to understand if perfomance critical cause when=expr looks pretty good from customization point

vkorolev: Yes, by perf reasons we'd better use something event-driven like 
`<q-grid:td on='edit'>`

It would be great, if we could check when=expr value manually without adding watchers. But I don't think, it's possible

docs/template.md
klumba12: but for example. i want to highlight row with particular data, how i can do it with css? without any dom manipulations, like adding class?

klumba12: another case is row details....

docs/template.md
klumba12: yes actually complicated, logic like in your case can be put inside of the template
but still maybe it will be usefull to have that kind of triggers
maybe we will support 2variants?
`<g-grid:td on:mode='edit'>` - doesn't require watch
`<g-grid:td on='$mode === edit'>` requires watch?

docs/template.md
unuscontraomnes: Why q? Is there some concept?

docs/template.md
vkorolev: We can provide some pre-defined css classes for different states of row.  Also for row customization we can make soemthing like this:
<q-grid:tr ng-class="{ extraordinaryClass: $row.condition }"></q-grid:tr>

klumba12: nice, i like it, actually we can interpolate all attributes to real tr

klumba12: for td it can work too...

docs/template.md
vkorolev: Yeah, I think we can have 2 ways of change detection: event-based and predicate-based.
1) `<q-grid:td on='edit'>` - similar to ng-switch, but invoked only on predefined events.
2)  `<q-grid:td when="$mode == 'edit'">` - bound to digest loop to track all changes.

I can see one problem with second option. Predicates are not guaranteed to be mutual exclusive. And we can have a case when more then one template is suitable.

PR #2, opened 2016-12-23
Update header.sticky.md

titovmx:
Really user should just set option sticky to true to switch on

### DIFF
--- a/docs/header.sticky.md
+++ b/docs/header.sticky.md
@@ -1,12 +1,35 @@
 # qgrid 
+
+## Option 1 - settings in template
+
 ## markup
 ```html
-<q-grid rows='data'></q-grid>
+<q-grid rows='data'>
+  <q-grid:head sticky="true"></q-grid:head>
+</q-grid>
 ```
+
+## Option 2 - settings in controller
+
+## markup
+```html
+<q-grid rows='data'
+        settings='settings'>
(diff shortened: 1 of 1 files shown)


docs/header.sticky.md
```diff
@@ -1,12 +1,35 @@
 # qgrid 
+
+## Option 1 - settings in template
```
titovmx, line 3: Will we put all setting to template or configure it from controller? In my mind it would be good in controller to keep markup more simple and readable. And use markup to customise templates only.
So I suggest 2 options below.

titovmx, line 3: Also I have concern about scrollable container. In current implementation [sticky-header](https://github.com/titov-max/sticky-header) has special directive ```sticky-viewer``` that should be set to external container to track scroll event. 

So now I'm not sure how we should watch this in qGrid. Probably it requires some improvement of sticky-header, search of external container. But I think, we shouldn't provide additional options for this. If you have any ideas about possible implementation please let me know.

vkorolev: As for me, the first option looks more preferable, because sticky is about look-and-feel only, there's no related logic from user's point of view. 
Also I think, header must be sticky by default, and only `sticky="false"` must be set explicitly by user to prevent its stickiness.

PR #3, opened 2016-12-23
Project starter structure

klumba12:
folder structure
webkit
webkit server
karma
eslint
babel


vkorolev:
Great job!
I'm just not sure if we need to have predefined styles for components themselves. We just need to define a set of elements and classes for components, and theming is end-user issue.
Set of elements and classes would allow us to create pluggable theming and different themes can be distributed with other packages.
What do you guys think about it?

titovmx:
@vkorolev, I think we should have some initial styles. First point is a quick start to work, if user would like powerful tool with minimal settings. Second, we'll need styles for demo purposes. And probably we'll need set some styles like flexible containers and others.

vkorolev:
Well, if we're going to use framework-dependent stuff like flex from ng-material, user will be forced to use ng-material and Material Design with minor CSS customization.

For quick start we can implement some external stylesheets. probably as an external module, so user will just need to add them to bower/npm config

klumba12:
We can grid system from ng material it totally depends on css

klumba12:
And i like themes maybe we ll do material theme in parallel it wiil give us opportunity to test customizability?

vkorolev:
Yes, we'll definitely have to implement a theme in parallel for testing in demo. The main point is to make it decoupled from q-grid itself.

klumba12:
Yes i agree even in demo we can have theme selector

titovmx:
No, I wouldn't like use for it additional dependencies. Flexboxes can be used without specific framework and I think we'll need for some simple standart stuff from the box. It'll define common behaviour: flexibility, poisitioning, whitespaces, etc.

Then we'll have themes as external modules. It is a good point, I absolutely agree with you.

klumba12:
for our default template we can use grid system from ng material, just by copy-paste few lines of style to our css :)

PR #9, opened 2016-12-27
Demo add json data

klumba12:


etokareva (approved):

scripts/tojson.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
titovmx, line 1: There is should be added 'node' to eslint envs

demo/sandbox/index.js
```diff
@@ -1,12 +1,19 @@
 'use strict';
 
-export default function Controller() {
-	this.rows = [{firstName: 'Alex', lastName: 'Pre'}, {firstName: 'Pavel', lastName: 'Q'}];
+Controller.$inject = ['$http'];
+export default function Controller($http) {
+	const ctrl = this;
+
+	this.rows = [];
```
titovmx: More common use ctrl variable instead of ```this``` everywhere

titovmx (approved):
I updated code by my own comments, so I haven't any remarks

PR #15, opened 2016-12-28
Spec example and configuration

klumba12:


src/services/utility.spec.js
```diff
@@ -0,0 +1,32 @@
...
+	function testObject() {
+		return {
+			foo: 1,
+			bar: {
+				baz: 'hello'
+			}
+		};
+	}
+
+	it('pass null should not change target', () => {
```
vkorolev: Consider to have more fluent statements: `it('should not change target when null is passed', ...` :)

etokareva (approved):

titovmx (approved):

vkorolev (approved):

PR #18, opened 2017-01-13
Grid model implementation

klumba12:
* Ugly for now
* Need to add tests


takesanocean (approved):
Not bad

src/core/services/event.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+export default class Event {
+	constructor(){
+		this.handlers = [];
```
klumba12, line 5: try to implement with weakrefs?

src/core/infrastructure/model.bind.js
```diff
@@ -0,0 +1,46 @@
...
+		this.off = noop;
+	}
+
+	bind(model, name) {
+		this.off();
+		const source = this.source;
+
+		if (model) {
+			this.off = model[name + 'Changed'].on(function (e) {
+				for (let key of Object.keys(e.changes)) {
```
klumba12: very similar to the model impl, make one code for both?

src/components/grid/grid.js
```diff
@@ -1,21 +1,46 @@
...
 	bindings: {
 		rows: '<',
-		columns: '<'
+		columns: '<',
+		model: '<'
 	}
 };
 
 Controller.$inject = [];
 function Controller() {
```
klumba12: Make base controller for sync this\scope and model?

src/core/services/utility.spec.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+import {isObject} from './utility';
+
+describe('utility', () => {
+
+	describe('isObject', () => {
+
+		it('should return false when passing number', () => {
+			assert.isFalse(isObject(12));
```
titovmx: Should we follow the assert or expect styles only in specs? Or it should be different depending on purposes?

src/core/services/event.js
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	on(f) {
+		const handlers = this.handlers;
+		handlers.push(f);
+		return function () {
```
titovmx: Can it be writed as arrow function?

demo/pages/model/index.js
```diff
@@ -0,0 +1,83 @@
...
+			title: 'Member Since'
+		}
+	];
+
+	ctrl.gridModel.data({
+		columns: columns
+	});
+
+	ctrl.gridModel
+		.selectionChanged
```
vkorolev, line 68: It breaks fluency.
How about this:
```javascript
ctrl.gridModel
	.selectionChanged(function (e) {
		...
	})
```
or even
```javascript

ctrl.gridModel
	.selection({
		change: function (e) {
			...
		}
	})
```

src/core/services/event.js
vkorolev: Can we also add some arrow function check to eslint rule?

demo/pages/model/index.js
```diff
@@ -0,0 +1,83 @@
...
+		},
+		{
+			key: 'contact.email.primary',
+			title: 'Primary Email',
+			value: item => item.contact.email[0]
+		},
+		{
+			key: 'likes',
+			title: 'Likes',
+			value: item => item.likes.join(', ')
```
vkorolev, line 55: Arrow function are so beautiful for such cases. Awesome

src/core/services/utility.spec.js
titovmx: e.g. it can be look so in BDD style
```expect(isObject(12)).to.be.false```

vkorolev: I like bdd-style more. Especially it can be seen for complex cases:
`expect(12).not.to.be.an.instanceof(Object)`
I think, in this case we'll use all power of Chai

demo/pages/model/index.js
klumba12: Yep =)

demo/pages/model/index.js
klumba12: Actualy any registered model will have event with 'Changed' postifx e.g. dataChanged, headChanged, selectionChanged.
Here's shown the way who user can use it.

klumba12: This event is used by base component to sync controller binding with model

klumba12: Defently we can add to model any type of expresssions of buty i believe, but it'll be just sugar baby

PR #19, opened 2017-01-14
Custom cell templates

klumba12:


demo/pages/cell/index.html
```diff
@@ -0,0 +1,11 @@
+<q-grid columns="$ctrl.columns"
+		  rows="$ctrl.rows">
+	<q-grid:body>
+		<q-grid:td key="name.last">
+			<b>{{$ctrl.$value()}}</b>
```
klumba12: need to get rid of $ctrl

src/core/infrastructure/guard.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+
+/**
+ * If value is null or undefined exception will be thrown
+ * @param {any} value - Value to check
+ * @param {string} name - Argument name
+ */
+export function notNull(value, name){
+	if (value === null || isUndefined(value)) {
```
titovmx, line 22: May adding isNull function make sense?

src/core/infrastructure/guard.js
```diff
@@ -0,0 +1,36 @@
+import Error from  './error';
+import {isUndefined} from '../services/utility';
+
```
titovmx, line 3: Can you show me where it's using?

klumba12: what do you mean? isUndefined used in guard e.g.

src/core/infrastructure/guard.js
klumba12: I'm not sure, we'll try to write in "not undedined" style, however maybe we can do it

src/core/infrastructure/guard.js
titovmx: I mean guard exactly :)

titovmx (approved):

PR #21, opened 2017-01-16
Make head templatable

klumba12:


src/components/body/body.template.js
```diff
@@ -0,0 +1,22 @@
...
+	constructor() {
+		super('body');
+	}
+}
+
+BodyTemplate.$inject = ['$element', '$transclude'];
+
+export default {
+	require: {
+		root: `^${GRID_NAME}`
```
titovmx, line 16: why do not use `**^^**${GRID_NAME}` ?

titovmx (approved):

src/components/body/body.template.js
klumba12: Body should be on the first level

PR #24, opened 2017-01-18
Demo improvements

klumba12:


dmitriy-kostianetskiy (approved):
It looks very good. I've not found any suspicious code. 🖒

PR #25, opened 2017-01-19
New template design

klumba12:


src/index.js
```diff
@@ -2,38 +2,38 @@ import angular from 'angular';
...
 	.service(def.SERVICE_NAME, () => () => new Model())
 	.run(Setup)
 	.name;
 
 Setup.$inject = ['$templateCache'];
 function Setup($templateCache) {
-	$templateCache.put('qgrid.html', require('./components/grid/grid.html'));
-	$templateCache.put('qgrid.head.cell.html', require('./components/head/cell.html'));
-	$templateCache.put('qgrid.body.cell.html', require('./components/body/cell.html'));
+	$templateCache.put('qgrid.tpl.html', require('./components/grid/grid.html'));
```
titovmx, line 36: tpl is a template? I don't like this short form.

klumba12: it's like common extension ('tpl.html'),
please suggest your naming:)

titovmx: why 'tpl' part is neccessary?
I would prefer 'template.html'

klumba12: still tpl is the common approach, think about it like about an extension

titovmx (approved):

PR #29, opened 2017-01-20
Export API template added

mikezotovdev:


docs/export.md
```diff
@@ -5,8 +5,74 @@
 ```
 ## script
 ```javascript
-function controller($scope){
-	$scope.data = [];
+//export API:
```
klumba12, line 8: how this works for the user? how this works for the model? how it will like for grid settings? =)

PR #32, opened 2017-01-20
Pinned column

etokareva:


docs/column.pin.md
```diff
@@ -1,12 +1,25 @@
 # qgrid 
 ## markup
 ```html
+<q-grid rows="$ctrl.rows">
+	<q-grid:columns>
+		<q-grid:column isPinned="true" key="name.last" title="Last Name" value="$ctrl.getLastName($row)"></q-grid:column>
```
vkorolev: `is-pinned` in this case :)
But as for me just `pinned` would be enough.

docs/column.pin.md
```diff
@@ -1,12 +1,25 @@
...
 ```
 ## script
 ```javascript
-function controller($scope){
-	$scope.data = [];
+controller.$inject = ['$scope', 'qgrid'];
+function controller($scope, qgrid){
+  $scope.columns = [
+          {key: 'firstName', label: 'First Name', type: 'text', isVisible: true, isDefault: true, isPinned: true},
```
vkorolev: Consider to use following default values: 
* visible - true
* default - true
* pinned - false

For other cases user will be add configuration

docs/column.pin.md
klumba12: yep, we'll make desicion to use or not to use 'is' :)

docs/column.pin.md
klumba12: default is reserved word, so maybe use 'is'?

klumba12: it's good practice to use is/can maybe for predicating...

docs/column.pin.md
klumba12: maybe pin='left|right' ?

docs/column.pin.md
klumba12: pin='left|right' ?

docs/column.pin.md
vkorolev: Yeah, i like this in markup

docs/column.pin.md
vkorolev: Yes, in JS-configuration we can use is- prefix, it looks good. pin='left|right' looks good too

etokareva: Maybe pin='left|right' for html and isPinned='left|right' for js then?

docs/column.pin.md
```diff
@@ -1,12 +1,25 @@
...
+controller.$inject = ['$scope', 'qgrid'];
+function controller($scope, qgrid){
+  $scope.columns = [
+          {key: 'firstName', label: 'First Name', type: 'text', isVisible: true, isDefault: true, isPinned: 'left'},
+          {key: 'lastName', label: 'Last Name', type: 'text', isVisible: true, isDefault: true, isPinned: 'left'},
+          {key: 'birthDate', label: 'Date of Birth', type: 'date', isVisible: true, isDefault: true},
+          {key: 'location', label: 'Location', type: 'date', isVisible: true, isDefault: true, isPinned: 'right'},
+          {key: 'zipCode', label: 'Zip', type: 'number', isVisible: false, isDefault: false, isPinned: 'right'}
+      ];
+
```
klumba12, line 23: Ispinned to pin

klumba12: Also label to title 😀

takesanocean: maybe not `pin="left"` but `pinned="left"`?
Sounds more natural to me.

klumba12: pinned it sounds like a boolean state for me :)
however we need to have common style pin vs pinned, canResize vs resizable, canMove vs moveable, isVisible vs visible, isDefault vs 'default' etc.

PR #41, opened 2017-01-23
Aggregation

mkoty:


docs/aggregation.md
```diff
@@ -1,12 +1,34 @@
...
 ```html
+<q-grid rows='data'>
+	<q-grid:columns>
+		<q-grid:column key="Name" title="Name" aggregation="count"></q-grid:column>
+		<q-grid:column key="Age" title="Age" aggregation="avg"></q-grid:column>
+		<!-- max || min || sum || count || avg -->
+	</q-grid:columns>
+	<!-- naming: aggregation, aggregation-footer -->
+	<q-grid:aggregation></q-grid:aggregation>
```
klumba12, line 11: right now we have for custom aggregation, one of the reasons to do like that - agg should be directly under the column values
```html
<q-grid columns="$ctrl.columns"
		  rows="$ctrl.rows">
	<q-grid:columns>
		<q-grid:column key="gender">
			<q-grid:template for="foot" let="$ctrl">
				<b>male: {{100*$ctrl.getCountGender($view.rows, 'male')/$view.rows.length}}%</b> <br/>
				<b>female: {{100*$ctrl.getCountGender($view.rows, 'female')/$view.rows.length}}%</b>
			</q-grid:template>
		</q-grid:column>
	</q-grid:columns>
</q-grid>
```

docs/aggregation.md
```diff
@@ -1,12 +1,34 @@
-# qgrid 
+# qgrid
 ## markup
 ```html
+<q-grid rows='data'>
+	<q-grid:columns>
+		<q-grid:column key="Name" title="Name" aggregation="count"></q-grid:column>
```
klumba12, line 6: so if aggregation is set for the column we show value in the footer for that column?

docs/aggregation.md
klumba12, line 11: i'm not sure about q-grid:aggregation component, guys what are yours thoughts?

docs/aggregation.md
```diff
@@ -1,12 +1,34 @@
...
+</q-grid>
 ```
 ## script
 ```javascript
-function controller($scope){
-	$scope.data = [];
+controller.$inject = ['$scope', 'qgrid'];
+function controller($scope, qgrid){
+	this.getColumns: function() {
```
klumba12, line 29: are we going to have service for js aggregations ?
like
```javascript
const srv = grid.service()
const cellValue = srv.aggregation.average(data)
```

docs/aggregation.md
vkorolev: I'm not sure about this. If we're talking about aggregation of the whole dataset (e.g. count or avg by all collection, we don't need any aggregation at all since we have `footer` component.
But if we need some `group by`-like functionality, we definitely need templates for aggregation

docs/aggregation.md
vkorolev: Does it work like `group by` ?

docs/aggregation.md
klumba12: Actually we are going to have 2 types of agg row and cilumn. Column in the footer, row when grouping will be applied

src/core/services/aggregation.js
```diff
@@ -0,0 +1,46 @@
+export default class Aggregation {
+	constructor() {
+		this.rows = null;
```
vkorolev: should `this` work with static methods?

klumba12 (approved):

PR #44, opened 2017-01-24
Checkable rows

dmitriy-kostianetskiy:


docs/row.checkable.md
```diff
@@ -1,24 +1,41 @@
...
+Checkable rows is just a one of the selection modes we offer to user.
+
+Here is a list of features we should support:
+* Extra column with checkboxes for each row
+* Select all checkbox in the header of the column
+* Column and header should be templatable
+* Selected items should be bound to `selection`
+
+When qgrid works in server side pagination mode we might face a trouble with `selection`.
+Just a simple case:
```
vkorolev: I think user has to take care of this. If user really needs to check all items in dataset, he should use client-side pagination

klumba12: This is complex behavior indeed

klumba12: But really we need to provide awesome API and user should decide what to do with that

klumba12: Also we do some work on groupbox directives please review it https://github.com/klumba12/groupbox

demo/pages/checkable-rows/index.html
```diff
@@ -0,0 +1,30 @@
+<q-grid columns="$ctrl.columns"
+		  rows="$ctrl.rows"
+		  selection="$ctrl.selection"
+		  selection-mode="$ctrl.selectionMode"
+		  on:selection-changed="$ctrl.selectionChanged($event)">
+	<q-grid:check>
```
klumba12: suggest:
```javascript
 <q-gird:selection>
     <q-grid:template for='select-all | select | cell | row | column'
```

klumba12: need to under, what is beeter select-all vs head select vs body select-all vs check-all vs checkbox-all vs box-all

demo/pages/checkable-rows/index.html
```diff
@@ -0,0 +1,30 @@
...
+				Select {{$row.name.last}}
+			</label>
+		</q-grid:template>
+	</q-grid:check>
+	<q-grid:head sticky="false"></q-grid:head>
+	<q-grid:toolbar>
+		<q-grid:template for="top" let="$ctrl">
+			<button ng-click="$ctrl.selectAll()">Select all</button>
+			<button ng-click="$ctrl.selectTop10()">Select top 10</button>
+			<button ng-click="$ctrl.diselectAll()">Diselect all</button>
```
klumba12: typo deselect

demo/pages/checkable-rows/index.js
```diff
@@ -0,0 +1,85 @@
...
+			title: 'Likes',
+			value: item => item.likes.join(', ')
+		},
+		{
+			key: 'memberSince',
+			title: 'Member Since'
+		}
+	];
+
+	this.selectionMode = 'check';
```
klumba12, line 62: guys any ideas for naming? check-row checkabke-row smth better please?

src/ng/components/selection/indeterminate.js
```diff
@@ -0,0 +1,28 @@
...
+	constructor($scope, $element, $attrs) {
+		super();
+
+		this.$scope = $scope;
+		this.$element = $element;
+		this.$attrs = $attrs;
+	}
+
+	onInit() {
+		this.$scope.$watch(this.$attrs[INDETERMINATE_NAME], (newVal) => {
```
klumba12, line 14: attrs has observe method, maybe better will be to use it

src/ng/components/selection/selection.core.js
```diff
@@ -0,0 +1,84 @@
+import Directive from '../directive';
+import {VIEW_CORE_NAME, SELECTION_CORE_NAME} from '../../../definition';
```
klumba12: use src/defenition core/infrastructure...

src/ng/components/selection/selection.core.js
```diff
@@ -0,0 +1,84 @@
...
+
+class SelectionCore extends Directive(SELECTION_CORE_NAME, {view: `^^${VIEW_CORE_NAME}`}) {
+	constructor($scope, $element, $attrs) {
+		super();
+
+		this.$scope = $scope;
+		this.$element = $element;
+		this.$attrs = $attrs;
+		
+		this.selectedItems = new Set();
```
klumba12, line 12: it looks like it should be part of selectionModel; selection model already contains that stuff

klumba12 (approved):

klumba12 (changes requested):
Accident approve

demo/pages/checkable-rows/index.html
dmitriy-kostianetskiy: I vote for select-all and select

vkorolev: select-all is good.
How about master?

demo/pages/checkable-rows/index.js
vkorolev: As I understood it's about radiobutton-mode and checkbox-mode?
How about multiple | single ?

src/ng/components/selection/indeterminate.js
dmitriy-kostianetskiy: $observe fires only if I change binding expression itself and newVal always has the same value "$selection.indeterminate()" in my case. Rather than $watch callback fires each time function result $selection.indeterminate() changes)

src/ng/components/selection/selection.core.js
dmitriy-kostianetskiy: Yes, you're right, we have an array of selected rows within our model. We also don't have any kind of isSelected property. Therefore we can't just bind ng-checked to the row's field. We need to bind to kind of a function which determines if row selected or not. Generally this is bad practice and it might dramatically impact performance, but anyway I tried to make state(row) function as fastest as possible using Set lookup instead of naive direct search through the items array within our selection model.

demo/pages/checkable-rows/index.js
dmitriy-kostianetskiy: I thought it's about **selection** mode such as checkbox(additional column with checkboxes), row selection(multiple or single) and particular cell selection :)

demo/pages/checkable-rows/index.html
```diff
@@ -0,0 +1,30 @@
+<q-grid columns="$ctrl.columns"
+		  rows="$ctrl.rows"
+		  selection="$ctrl.selection"
+		  selection-mode="$ctrl.selectionMode"
+		  on:selection-changed="$ctrl.selectionChanged($event)">
+	<q-grid:selection>
+		<q-grid:template for="select-all">
```
klumba12, line 7: Looks beautifull!!!

demo/pages/checkable-rows/index.html
```diff
@@ -0,0 +1,30 @@
...
+			<label>
+				<input type="checkbox" q-grid:indeterminate="$selection.indeterminate()" 
+					ng-checked="$selection.state()" 
+					ng-click="$selection.toggleAll()"/>
```
klumba12, line 11: we began to use command pattern $selection.toggleAll.(canExecute|execute) for acion templating.
see ng/plugin/groupbar.js

demo/pages/checkable-rows/index.html
```diff
@@ -0,0 +1,30 @@
+<q-grid columns="$ctrl.columns"
+		  rows="$ctrl.rows"
+		  selection="$ctrl.selection"
+		  selection-mode="$ctrl.selectionMode"
+		  on:selection-changed="$ctrl.selectionChanged($event)">
+	<q-grid:selection>
+		<q-grid:template for="select-all">
+			<label>
+				<input type="checkbox" q-grid:indeterminate="$selection.indeterminate()" 
+					ng-checked="$selection.state()"
```
klumba12, line 10: for indeterminate and checked here can we use getters?
ng-checked="$selection.state" ? or it's done for compatibility with select template state($row)?

demo/pages/checkable-rows/index.js
klumba12: now i see next selection.modes =  row, cell, column, range
maybe add 
selection.check = true,
selection.checkMode = 'single|multiple'}
so in html all selection go to selection component?
```html
<qgrid>
   <qgrid:selection items="$ctrl.mySelections"
                              mode="row"
                              check|can-check="true"
                              check-mode="single">
</qgird
```

klumba12: Or i'm going to introduce global visibilty model...maybe there...

src/ng/components/selection/selection.core.js
klumba12: Ok maybe then rename it to selectionSet? just to avoid confustion?)

src/ng/components/selection/selection.core.js
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		return this.selectedItems.has(row);
+	}
+
+	indeterminate() {
+		return this.selectedItems.size !== this.view.rows.length && this.selectedItems.size > 0;
+	}
+
+	updateSelection() {
```
klumba12, line 41: can be called invalidate

src/ng/components/selection/selection.core.js
```diff
@@ -0,0 +1,81 @@
...
+		const selection = this.view.model.selection;
+
+		selection({
+			items: Array.from(this.selectedItems)
+		});
+	}
+
+	toggle(row) {
```
klumba12, line 49: As i wrote above for user accessable actions consider to use new Command...Thanks!

klumba12 (approved):

PR #49, opened 2017-01-25
Sticky fixes and improvements.

titovmx:
- sticky was subscribed on theme change
- dom observe functions was placed to own service


src/ng/components/sticky/sticky.core.js
```diff
@@ -44,13 +45,11 @@ class StickyCore extends Directive(STICKY_CORE_NAME, {view: `^^${VIEW_CORE_NAME}
...
 
-		sticky.invalidated.on(() => {
-			self.$scope.$apply();
-		});
+		this.view.theme.changed.on(() => self.$timeout(() => sticky.invalidate()));
 
-		this.$scope.$on('$destroy', () => {
-			sticky.destroy();
-		});
+		sticky.invalidated.on(() => self.$scope.$apply());
```
vkorolev: Maybe `sticky.invalidate.on` ?

titovmx: Sticky has already method ```invalidate```. So ```invalidated``` will be an event when invalidation finish.

klumba12: +1

src/core/services/observe.js
```diff
@@ -0,0 +1,53 @@
+export function children(element, handler) {
```
klumba12: maybe dom.observe is better name. it indicates that observations can occur only on dom elemeents

src/core/services/observe.js
```diff
@@ -0,0 +1,53 @@
...
+	const observer = new MutationObserver((mutations) => {
+		mutations.forEach((mutation) => {
```
klumba12: can remove parenthesis mutation =>

src/core/services/observe.js
klumba12: what do you think about next variants:
```javascript
 observe.children(element, e => ) //current
 observe.children(element).on(e =>
 observe(element).children.on(e => ...)
 observe(element).children(e =>
 observe(element).childrenChanged.on(e =>
```

klumba12 (approved):

PR #51, opened 2017-01-26
Plugin model

klumba12:


docs/sandbox.md
```diff
@@ -2,103 +2,78 @@
...
-    $scope.onColumnOrderChanged = function(e){
-    };
+controller.$inject = ['qgrid', '$http'];
+function controller(qgrid, $http){
+	const gridModel = qgrid.model();
+	const gridService = qgrid.service(gridModel);
+	
+	gridModel.data({
+		pipe: [
+				(data, model, next) => {
```
vkorolev, line 66: Context instead of model, I guarantee :)

src/core/pagination/pagination.model.js
```diff
@@ -0,0 +1,9 @@
+import Resource from '../resource/resource';
+
+export default class PaginationModel {
+	constructor() {
+		this.resource = new Resource();
```
vkorolev, line 5: Don't we need a current page as well?

src/core/services/utility.js
```diff
@@ -26,6 +28,9 @@ export {
 	isUndefined,
 	clone,
 	debounce,
+	merge,
+	yes,
```
vkorolev, line 32: How about `truthy`|`falsy` or even `true` | `false` ?
Similar naming is used in assertion libs

src/core/pagination/pagination.model.js
klumba12: current will be a part of scroll. maybe view, maybe of pagination, so for now let it be fake current)

src/core/services/utility.js
klumba12: true|false is system stuff don't wont to mix it, truthy vs yes can make vote in the future)

docs/sandbox.md
klumba12: =) yep) it'll continously updated bro =)

vkorolev (approved):

PR #56, opened 2017-01-27
Simple groupibar UI

klumba12:


src/ng/components/template/template.path.js
```diff
@@ -0,0 +1,55 @@
...
+				`"${name}" is already registered`);
+		}
+
+
+		resolvers[name] = resolve;
+		return TemplatePath;
+	}
+
+	static get(source) {
+		const path = this.find(source);
```
vkorolev, line 23: All methods are static. Does `this` work with static context?

klumba12: they unfortunate static for now, because i couldnt come up how to require it to controller differently, i think it'll be evolved

vkorolev (approved):

src/ng/plugins/groupbar/groupbar.js
```diff
@@ -0,0 +1,61 @@
...
+		this.newGroup = null;
+		this.add = new Command({
+				execute: key => {
+					const group = this.model.group;
+					const state = group();
+					group({
+						by: state.by.concat(key)
+					});
+
+					this.newGroup = null;
```
titovmx, line 17: key and this.newGroup are both the same?

PR #62, opened 2017-01-27
Aggregation update

mkoty:


src/core/services/aggregation.js
```diff
@@ -1,34 +1,66 @@
...
 	static max(rows, getValue) {
-		if(!rows.length){
+		if (!rows.length) {
 			return null;
 		}
 
-		var max = Math.max.apply(Math, rows.map(getValue));
-		return max;
+		return getValue(_.maxBy(rows, d => getValue(d)));
+	}
```
klumba12: 1. _.maxBy(rows, getValue) - is enough
2. Let's write our own super fast version with while(true), it's easy and agg perfomance is critical operation for grouping aggregations

src/core/services/aggregation.js
```diff
@@ -1,34 +1,66 @@
...
+		if (!rows.length) {
+			return null;
+		}
+
+		return getValue(_.minBy(rows, d => getValue(d)));
+	}
+
+	static minMax(rows, getValue) {
+		return [this.min(rows, getValue), this.max(rows, getValue)];
+	}
```
klumba12: need to rewrite it to one cycle we goes 2 times throught all rows here

src/core/services/aggregation.js
```diff
@@ -1,34 +1,66 @@
...
+
+	static avg(rows, getValue) {
+		const length = rows.length;
+		return rows.reduce(function (reduced, d) {
+			return reduced + getValue(d) / length;
+		}, 0);
+	}
```
klumba12: rewrite to while

src/core/services/aggregation.js
```diff
@@ -1,34 +1,66 @@
...
+		return rows.reduce(function (reduced, d) {
+			return reduced + Number(getValue(d));
+		}, 0);
+	}
+
+	static join(rows, getValue) {
+		return rows.reduce(function (reduced, d) {
+			return reduced + getValue(d).toString();
+		}, '');
 	}
```
klumba12: rewrite it to while check out what is faster getValue(d).toString() or ''+getValue(d)

src/core/services/aggregation.js
```diff
@@ -1,34 +1,66 @@
...
+		const length = rows.length;
+		return rows.reduce(function (reduced, d) {
+			return reduced + getValue(d) / length;
+		}, 0);
+	}
+
+	static sum(rows, getValue) {
+		return rows.reduce(function (reduced, d) {
+			return reduced + Number(getValue(d));
+		}, 0);
```
klumba12: why somewhere Number() is used and somewhere not?

mkoty: On sum Number needed (because without it sum concat strings to one), and on avg it's not need because automatic cast occured

klumba12 (approved):

PR #64, opened 2017-01-28
Expandable groups

klumba12:


src/ng/services/value.js
```diff
@@ -1,12 +1,16 @@
+import {isFunction} from 'core/services/utility';
+
 export function get(row, column) {
-	return column.$value
-		? column.$value({$row: row})
-		: column.value
-			? column.value(row)
-			: row[column.key];
+	return row.value && isFunction(row.value)
```
klumba12, line 4: for perfomance reasons...it'll we refactored to factories later i guess

vkorolev (approved):

PR #66, opened 2017-01-29
Data pipeline implementation

vkorolev:
Initial implementation of Pipeline.


src/core/pipeline/pipe.js
```diff
@@ -0,0 +1,35 @@
+import * as guard from '../infrastructure/guard';
+
+export default class Pipe {
```
vkorolev: Additional function will be added: order, group etc

src/core/pipeline/pipeline.js
```diff
@@ -0,0 +1,35 @@
+export default class Pipeline {
```
vkorolev: Here's brief list of what needs to be done. Guys, please have a look at it
1. Pipeline's stage exception handling
2. Timeouts?
3. Debounce?

src/core/infrastructure/guard.js
```diff
@@ -33,4 +33,15 @@ export function notNullOrEmpty(value, name){
...
 		throw new AppError('guard.notNullOrEmpty', name);
 	}
+}
+
+/**
+ * If value is not a function exception will be thrown
+ * @param {any} value - Value to check
+ * @param {string} name - Argument name
+ */
+export function invokable(value, name){
```
klumba12, line 43: maybe just guard.function ?

src/core/pipeline/pipe.js
```diff
@@ -0,0 +1,35 @@
+import * as guard from '../infrastructure/guard';
```
klumba12: Can we call folder also pipe?
and each pipe has own file?
pipe.filter.js
pipe.pagination.js
etc ...

src/core/pipeline/pipe.js
```diff
@@ -0,0 +1,35 @@
...
+export default class Pipe {
+	static get filter() {
+		return (data, ctx, next) => {
+			let model = ctx.model;
+			guard.notUndefined(model.filter, 'model.filter');
+
+			let filter = model.filter();
+			guard.invokable(filter.match, 'model.filter.match');
+
+			let result = [];
```
klumba12: can be declare as const result

src/core/pipeline/pipe.js
```diff
@@ -0,0 +1,35 @@
+import * as guard from '../infrastructure/guard';
+
+export default class Pipe {
+	static get filter() {
+		return (data, ctx, next) => {
+			let model = ctx.model;
+			guard.notUndefined(model.filter, 'model.filter');
```
klumba12: you don't need to check that it'll be core model

src/core/pipeline/pipe.js
```diff
@@ -0,0 +1,35 @@
...
+
+			let filter = model.filter();
+			guard.invokable(filter.match, 'model.filter.match');
+
+			let result = [];
+
+			for (let i = 0, length = data.length; i < length; i++) {
```
klumba12: next step maybe: to use while loop for perf. critical operations :) while(--length?

src/core/pipeline/pipe.js
```diff
@@ -0,0 +1,35 @@
+import * as guard from '../infrastructure/guard';
+
+export default class Pipe {
+	static get filter() {
+		return (data, ctx, next) => {
+			let model = ctx.model;
+			guard.notUndefined(model.filter, 'model.filter');
+
+			let filter = model.filter();
+			guard.invokable(filter.match, 'model.filter.match');
```
klumba12: most common solution will be:
const match = filter.match(ctx.model)
so filter.match is a factory...

src/core/pipeline/pipeline.js
klumba12: better name. pipe.js pipeline.js: 
My suggestions:
pipe.build.js
pipe.run.js
pipe.middleware (the best for me)
pipe.request

klumba12: 1. Not a middleware work
2. Not a middleware work
3. Not a middleware work :)

src/core/pipeline/pipeline.js
```diff
@@ -0,0 +1,35 @@
...
+
+	constructor(stages) {
+		this.stages = stages || [];
+	}
+
+	fetch(ctx, data = []) {
```
klumba12: run is better name (from asp middleware)

src/core/pipeline/pipeline.js
```diff
@@ -0,0 +1,35 @@
...
+		return pipeline(
+			this.stages.map(s => data => new Promise(resolve => {
+				s(data, ctx, resolve);
+			})),
+			data);
+	}
+}
+
+function pipeline(stages, initialData) {
```
klumba12: looks overcomplicated:
instead of class can have 1 function

```javascript
run(stages, data, context){
   if(stages.length){
      stage[0](data, context, data => run(stages.slice(1), data, context) 
   }
}
```
with closing context:
```javascript
build(context){
   return function next(stages, data){
      if(stages.length){
         stage[0](data, context, data => next(stages.slice(1), data)
      }   
   }
}
const run = build(context);
run(stages, somedata)
```
the 2nd is my favor :)

src/core/infrastructure/guard.js
vkorolev: Yeah, I thought about that, but `export function function`. Probably I'll rewrite all exports

src/core/pipeline/pipe.js
vkorolev: Good idea

src/core/pipeline/pipeline.js
vkorolev: 1. What if exception happens in user-defined stage? We should catch that stuff and stop the pipeline i guess.

2 and 3 - Probably it's not directly middleware's work. But probably we need a way to stop the middleware outside?

src/core/pipeline/pipeline.js
vkorolev:
```javascript
pipe
    .run()
    .then((data) => ...)
```

In context of grid pipeline performs a kind of `read` operation. As for me `run` could be good for a generic middleware that doesn't return anything.

We can implement generic async middleware with `run` and then invoke `pipe.read` which internally will use `middleware.run`

src/core/pipeline/pipeline.js
klumba12: 1. I'm not sure how you will do it for async operations, but defenetly we can put try catch, and invoke the last(system finish) pipe?
2. You mean cancel, good point? if we are going to provide cancelation in context, next shoud be there to
(data, ctx) => if something wrokng { ctx.cancel() ??

src/core/infrastructure/guard.js
klumba12: yeees, maybe func will works too or not?

src/core/pipeline/pipeline.js
klumba12: why? I think promises here is overcomplicated, what is usecase for them? Maybe we can return Promise in invalidate, but in middelware?
```javascript
qgridService.invalidate => (){
 const promise...
 const pipe = [start].concat(model.pipe).concat([finish, () => promise.resolve())
  const run = build(new PipeContext(model));
  run(pipe);
  return promise...
}
```

src/core/pipe/pipe.sort.js
```diff
@@ -0,0 +1,27 @@
...
+	const values = [],
+		directions = [],
+		model = ctx.model,
+		dataModel = model.data(),
+		columns = dataModel.columns,
+		sort = model.sort(),
+		by = sort.by;
+
+	for (let i = 0, length = by.length; i < length; i++) {
+		let item = by[i],
```
klumba12: Use sort.service code

src/core/column/column.service.js
```diff
@@ -1,6 +1,14 @@
+import {isFunction} from '../services/utility';
+
 export function map(columns) {
 	return columns.reduce((memo, column) => {
 		memo[column.key] = column;
 		return memo;
 	}, {});
+}
+
+export function getValue(column) {
```
klumba12: This is more complicated function look at value.js

src/core/pipe/pipe.filter.js
```diff
@@ -0,0 +1,21 @@
+import * as guard from '../infrastructure/guard';
+
+export default function (data, ctx, next) {
+	const model = ctx.model;
+	guard.notUndefined(model.filter, 'model.filter');
```
klumba12: don't need guard here it, it's core model

src/core/pipe/pipe.filter.js
```diff
@@ -0,0 +1,21 @@
...
+
+	const filter = model.filter();
+	guard.invokable(filter.match, 'model.filter.match');
```
klumba12: i think we don't need to guard it like that, need to think about model type persisting...

src/core/pipe/pipe.middleware.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+function pipeline(stages, initialData) {
+	let index = 0;
+
+	return new Promise((resolve, reject) => {
+		invoke(initialData);
+
+		function invoke(data) {
+			let stage = stages[index++];
```
klumba12: can be that stages will mutate during invokation? maybe we need to make copy of it?

src/core/pipe/pipe.middleware.js
```diff
@@ -0,0 +1,37 @@
...
+
+			p.then(function (data) {
+				if (index < stages.length) {
+					invoke(data);
+				} else {
+					resolve(data);
+				}
+			}).catch((e) => {
+				reject(e);
+			});
```
klumba12: Cool!

src/core/pipe/pipe.middleware.js
```diff
@@ -0,0 +1,37 @@
...
+	run(ctx, data = []) {
+		return pipeline(
+			this.stages.map(s => data => new Promise(resolve => {
+				s(data, ctx, resolve);
+			})),
+			data);
+	}
+}
+
+function pipeline(stages, initialData) {
```
klumba12: can we reaname initialData to rows? so it'll show what type of data we are expecting on the start?)

src/core/pipe/pipe.filter.js
vkorolev: I think this guard is needed because for core model `filter` property is not required is it?

User can use `pipe.filter` function only if he defined `model.filter`.

Unlike that `guard.function` placed below is needed to be check inside the model, i agree :)

src/core/column/column.service.js
titovmx: @klumba12, can value.js be placed to core?

src/core/pipe/pipe.sort.spec.js
```diff
@@ -0,0 +1,104 @@
+import sort from './pipe.sort';
+
+describe('sort pipe', () => {
+	const list = [{
```
titovmx, line 4: May be there would be better to set source list in more "shaked" state. Actually first test doesn't change order for asc sorting by name.

vkorolev: Good point

src/core/column/column.service.js
vkorolev: Yeah, it would be good to keep such stuff in core.
But isn't value.js related to angular-specific model? There's something like `column.$value`

src/core/pipe/pipe.middleware.js
vkorolev: Yeah, good point. But in `run` method there's `this.stages.map` which is similar to cloning

src/core/pipe/pipe.filter.js
klumba12: no it's required, we only should care about external plugin model and that all

src/core/services/grid.js
```diff
@@ -1,9 +1,20 @@
...
+		const model = this.model;
+		const stages = model.data().pipe;
+		const ctx = {
+			model: model,
+			source: 'invalidate'
+		};
+		const middleware = new Middleware(stages);
+		middleware
+			.run(ctx, model.data().rows)
+			.then(result => model.data({rows: result}));
```
vkorolev: Not sure, what is going to be in `catch` block. Probably we need to have a kind of `log` wrapper similar to angular `$log` service

src/core/pipe/pipe.sort.js
```diff
@@ -0,0 +1,29 @@
+import {getValue} from '../column/column.service';
+import {orderBy} from '../services/utility';
+import {key as getKey, direction as getDirection} from '../sort/sort.service';
+import {find} from '../column/column.service';
+
+export default function (data, ctx, next) {
+	const values = [],
+		directions = [],
+		model = ctx.model,
+		dataModel = model.data(),
```
klumba12: every line should consist of const
const values
const directions..
please check this style in other files too

klumba12: dataState = model.data() is became common naming convension

src/core/services/grid.js
```diff
@@ -1,9 +1,20 @@
...
 export default class GridService {
 	constructor(model) {
 		this.model = model;
 	}
 
 	invalidate() {
-
+		const model = this.model;
+		const stages = model.data().pipe;
+		const ctx = {
```
klumba12: use full name context, ctx is for user defined

src/core/services/utility.js
```diff
@@ -36,5 +37,6 @@ export {
 	yes,
 	no,
 	toCamelCase,
-	noop
+	noop,
+	orderBy
```
klumba12, line 41: can we make a part of sort.service?

src/core/column/column.service.js
```diff
@@ -1,10 +1,28 @@
...
+		: row => row[column.key];
+}
+
+export function find(columns, key) {
+	for (let i = 0, length = columns.length; i < length; i++) {
+		if (columns[i].key == key) {
+			return columns[i];
+		}
+	}
+	throw new AppError('Not Found', `column ${key} is not found`);
```
klumba12: first arg should be column.service, Column should be in upper case, $key decorated with double quotes "
also when call find doesn't expect exception can you return null, and throw outside?

src/core/column/column.service.js
```diff
@@ -1,10 +1,28 @@
...
 
+export function getValue(column) {
+	return isFunction(column.value)
+		? row => column.value(row)
+		: row => row[column.key];
+}
+
+export function find(columns, key) {
+	for (let i = 0, length = columns.length; i < length; i++) {
+		if (columns[i].key == key) {
```
klumba12: columns.findIndex?
or if for pefomance const column = columns[i];

src/core/column/column.service.js
```diff
@@ -1,10 +1,28 @@
...
+import AppError from '../infrastructure/error';
+
 export function map(columns) {
 	return columns.reduce((memo, column) => {
 		memo[column.key] = column;
 		return memo;
 	}, {});
 }
 
+export function getValue(column) {
```
klumba12, line 10: as i wrote this shouldn't be here, you need to export it to pipe context as valueFactory from value.js

src/ng/components/grid/grid.js
```diff
@@ -1,17 +1,21 @@
 import RootComponent from '../root.component';
 
 export class Grid extends RootComponent {
-	constructor($element, $transclude) {
+	constructor($element, $transclude, qgrid) {
```
klumba12, line 4: are u sure you need to use here ng service?

klumba12 (approved):

PR #67, opened 2017-01-29
Column drag and drop

klumba12:


src/ng/components/dnd/drop.js
```diff
@@ -0,0 +1,93 @@
...
+
+		this.$scope = $scope;
+		this.element = $element[0];
+	}
+
+	onInit() {
+		const element = this.element;
+
+		element.classList.add('can-drop');
+		element.addEventListener('dragenter', this.enter.bind(this), false);
```
vkorolev: `() => this.enter` should work

src/ng/components/dnd/drop.js
```diff
@@ -0,0 +1,88 @@
...
+		super();
+
+		this.$scope = $scope;
+		this.element = $element[0];
+		this.listener = new EventListener(this, this.element);
+	}
+
+	onInit() {
+		this.element.classList.add('can-drop');
+		this.listener.on('dragenter', this.enter);
```
vkorolev, line 17: won't `this` pointer be lost?
How about `this.listener.on('dragenter', e => this.enter(e))` ?

klumba12: no look at this.listener = new EventListener(THIS, element)
this is sepcial abstraction to symplify life =)

vkorolev (approved):

PR #71, opened 2017-01-31
Template refactoring

klumba12:


takesanocean (approved):
Looks like the new model is more convenient

PR #75, opened 2017-02-01
Fix pivot

klumba12:


titovmx (approved):
New concept actually looks better for extension and I like that more logic keep in core now. It would be great add specs for that and docs, it'll help other team members join to work on templates and next features :)

PR #78, opened 2017-02-01
Toolbar right (and left), visibility change

etokareva:


src/ng/plugins/visibility/visibility.html
```diff
@@ -1 +1,18 @@
-<!--<span>{{$visibility.model.visibility()}}</span>-->
\ No newline at end of file
+<h2>Visibillity: </h2>
```
klumba12: can we go throught visibility keys and not hard code it?
For hiearachy like toolbar just make structure like that
*head
*body
*toolbar
***top
***left

src/ng/components/view/view.html
```diff
@@ -1,9 +1,8 @@
-<q-grid-core:toolbar position="right">
-</q-grid-core:toolbar>
-<q-grid-core:toolbar position="top">
+<q-grid-core:toolbar ng-if="$view.visibility.toolbar.top"
+                     position="top">
 </q-grid-core:toolbar>
 <div q-grid-core:viewport
-     class="viewport">
+     class="viewport" ng-class="{'viewport-inline': $view.visibility.toolbar.right}">
```
klumba12: can we apply some magic css to avoid that?

src/ng/plugins/visibility/visibility.html
```diff
@@ -1 +1,16 @@
-<!--<span>{{$visibility.model.visibility()}}</span>-->
\ No newline at end of file
+<b>Visibillity: </b>
+<ul>
+	<li ng-repeat="(key, value) in $visibility.model.visibility() track by $index" ng-if="key !== 'resource'">
```
klumba12: Should be something like $visibility.items user who will overidde template shouldnt think about filtration

klumba12: And model shouldnt be propogated to template too 😀

src/ng/plugins/visibility/visibility.html
```diff
@@ -1 +1,16 @@
...
+		<input type="checkbox" ng-model="$visibility.model.visibility()[key]"
+		       ng-change="$visibility.change.execute($visibility.model.visibility()[key])"
+		       ng-if="key !== 'toolbar'"/>
+		<label>{{key}}</label>
+		<ul ng-if="key === 'toolbar'">
```
klumba12: Tree structure should be exposed from the controller not hrdcoded toolbar

src/ng/plugins/visibility/visibility.html
```diff
@@ -1 +1,16 @@
...
+			<li ng-repeat="(prop, values) in $visibility.model.visibility().toolbar track by $index">
+				<input type="checkbox" ng-model="$visibility.model.visibility().toolbar[prop]"
+				       ng-change="$visibility.change.execute($visibility.model.visibility().toolbar[prop])"/>
```
klumba12: Toggle is a comon name not change

src/ng/components/view/view.html
```diff
@@ -1,25 +1,32 @@
...
-		</thead>
-		<tfoot q-grid-core:foot
-		       q-grid-core:sticky
-		       ng-include="$view.templateUrl('foot')">
-		</tfoot>
-		<tbody q-grid-core:body
-		       ng-include="$view.templateUrl('body')">
-		</tbody>
-	</table>
+<div class="wrapper">
```
klumba12: Should be more meaningfull name

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,8 +14,16 @@ TemplatePath
 class Visibility extends PluginComponent('qgrid.plugins.visibility.tpl.html') {
 	constructor() {
 		super(...arguments);
+		this.toggle = new Command({
+			execute: key => true,
```
klumba12: key => true, can be changed to import {yes} from 'core/services/utility' :)

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,8 +14,16 @@ TemplatePath
...
+			canExecute: () => true
+		});
 
+	}
 
+	get items() {
+		let data = this.model.visibility();
```
klumba12: if variable is not reassignet use const
rename data to visbilityState, it became common name pattern

klumba12: need to use smth like
```javascript
build(graph){
const keys =...
const nodes = [];
for(key in...{
   node = new Node(key)
   if(isObject...{
     node.children(build(...
   }
  nodes.push(node)
}
return nodes;
}
```

src/ng/components/view/view.html
```diff
@@ -1,25 +1,32 @@
...
-		</thead>
-		<tfoot q-grid-core:foot
-		       q-grid-core:sticky
-		       ng-include="$view.templateUrl('foot')">
-		</tfoot>
-		<tbody q-grid-core:body
-		       ng-include="$view.templateUrl('body')">
-		</tbody>
-	</table>
+<div class="flexible">
```
klumba12: still not good name, it shows how it layouts internally and not what is that for...btw. wrapper was beetr :D
consider maybe dock-center, if so wrap top, bottom toolbars to div with classes dock-top and dock-bottom 
dock, and right left to dock-left and dock-right

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,15 +15,49 @@ TemplatePath
...
 
+	get items() {
+		let data = this.model.visibility(),
+			items = Object.assign({}, data);
+		for (let [key, value] of Object.entries(items)) {
+			if (key === 'resource') {
+				delete items[key];
+			}
+		}
+		console.log(this.iterate(items));
```
etokareva: Yes, I need to remove it :1st_place_medal:

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,15 +15,49 @@ TemplatePath
 class Visibility extends PluginComponent('qgrid.plugins.visibility.tpl.html') {
 	constructor() {
 		super(...arguments);
+		this.toggle = new Command({
+			execute: (key, parentKey = null) => {
+				if (parentKey) {
```
klumba12: it won't work with more complex objects
you need to close in Node setter/getter for visibility

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,15 +15,49 @@ TemplatePath
...
+					this.model.visibility()[key] = !this.model.visibility()[key];
+				}
+			},
+			canExecute: (key) => typeof this.model.visibility()[key] != "object"
+		});
+		this.iterate = (obj) => {
+			let nodes = [];
+			for (let [property, value] of Object.entries(obj)) {
+				let node = new Node(property);
+				if (typeof obj[property] == "object") {
```
klumba12: use isObject from utility

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,15 +15,49 @@ TemplatePath
...
+			execute: (key, parentKey = null) => {
+				if (parentKey) {
+					this.model.visibility()[parentKey][key] = !this.model.visibility()[parentKey][key];
+				} else {
+					this.model.visibility()[key] = !this.model.visibility()[key];
+				}
+			},
+			canExecute: (key) => typeof this.model.visibility()[key] != "object"
+		});
+		this.iterate = (obj) => {
```
klumba12: can you reneame it to build? iterate is not common, and obj to graph

src/ng/plugins/visibility/visibility.js
```diff
@@ -14,15 +15,49 @@ TemplatePath
...
+			let nodes = [];
+			for (let [property, value] of Object.entries(obj)) {
```
klumba12: property -> key, property oftern mean getter/setter

src/ng/plugins/visibility/visibility.html
```diff
@@ -1 +1,19 @@
-<!--<span>{{$visibility.model.visibility()}}</span>-->
\ No newline at end of file
+<b>Visibillity:</b>
+<ul>
+	<li ng-repeat="item in $visibility.items">
```
etokareva, line 3: I want to create recursive component to fit cases of different layers. I.e.,
`export default ListCore.component('list', {
	controller: ListCore,
	controllerAs: '$list',
	template: '<ul><li ng-repeat="node in $list.items"><input type="checkbox" ng-if="node.children.length"/><label>{{item.key}}</label><list tree="node.children"></list></li></ul>',
	bindings: {
		tree: '='
	}
}) 
`
to change the visibility template to `<list tree="$visibility.items"></list>`

But I don't know whether my guesses are correct and where can I put the component exactly.

klumba12 (approved):

PR #79, opened 2017-02-02
Template cache

klumba12:


src/core/infrastructure/cache.js
```diff
@@ -0,0 +1,50 @@
...
+
+		return entry;
+	}
+
+	has(key) {
+		const items = this.items;
+		return items.hasOwnProperty(key);
+	}
+
+	find(key) {
```
titov-max: Should ```find(key)``` be opened for user? Looks like unsafe ```get(key)``` :)

takesanocean (approved):

PR #80, opened 2017-02-02
Cell editor

klumba12:


src/ng/components/body/body.core.js
```diff
@@ -1,16 +1,61 @@
...
+		this.listener = new EventListener(this, this.element);
+		Object.defineProperty($scope, '$view', {get: () => this.view});
+
+		this.cellEdit = new Command({
+			canExecute: cell => {
+				const model = this.view.model;
+				if (model.edit().mode === 'cell') {
+					const columnMap = columnService.map(model.data().columns);
+					if (columnMap.hasOwnProperty(cell.column.key)) {
+						// use shouldn't explicitly set it in the template, cause we have here canEdit !== false
```
titovmx: but canEdit had already set to true by default, so why we can't return just ```cell.column.canEdit```?

titovmx (approved):

PR #84, opened 2017-02-02
Add Path Attribute to column

inyryaev:


src/ng/services/value.js
```diff
@@ -1,7 +1,17 @@
+import angular from 'angular';
+
 export function get(row, column) {
-	return column.$value
-		? column.$value({$row: row})
-		: column.value
+	if (column.path) {
+		var result = row;
+
+		angular.forEach(column.path.split('.'), function (value) {
```
klumba12, line 7: avoid to use angular... use js native methods

src/ng/services/value.js
```diff
@@ -1,7 +1,17 @@
+import angular from 'angular';
+
 export function get(row, column) {
-	return column.$value
```
klumba12, line 2: should support $value also
it has more priority then path if it set

src/ng/services/value.js
```diff
@@ -1,7 +1,17 @@
...
+			result = result[value];
+		})
+
+		return result;
+	}
+
+	return column.value
 			? column.value(row)
 			: row[column.key];
 }
```
klumba12, line 17: do the same for getFactory, you need to build getter that works really quick

src/ng/services/value.js
```diff
@@ -1,7 +1,17 @@
+import angular from 'angular';
```
klumba12, line 1: Maxim added path service to the core, need to reuse it

PR #86, opened 2017-02-02
Column auto generation

titovmx:


src/ng/components/column/column.list.js
```diff
@@ -1,12 +1,34 @@
...
+		const data = model.data;
+		const state = data();
+		
+		if (state.columns.length === 0 && this.generate) {
+			model.dataChanged.on(e => {
+				if (e.changes.hasOwnProperty('rows')) {
+					const deep = this.generateMode === 'deep' || isUndefined(this.generateMode);
+					if (!deep && this.generateMode !== 'flat') {
+						throw new AppError(
+							'template.core',
```
klumba12: txception has invalid message and title

src/ng/components/column/column.list.js
```diff
@@ -1,12 +1,34 @@
...
 
 	onInit() {
+		const model = this.root.model;
+		const data = model.data;
+		const state = data();
+		
+		if (state.columns.length === 0 && this.generate) {
+			model.dataChanged.on(e => {
+				if (e.changes.hasOwnProperty('rows')) {
+					const deep = this.generateMode === 'deep' || isUndefined(this.generateMode);
```
klumba12: use switch it more readable

src/core/column-list/generate.js
```diff
@@ -0,0 +1,54 @@
...
+						});
+						return accessor[prop].join('; ');
+					}
+				});
+			} else {
+				const mergeColumns = generate([source[prop]], true, key);
+				acc.push(...mergeColumns);
+			}
+		} else {
+			const parts = path !== '' ? path.split('.') : [];
```
klumba12: make a path service
const getSet|value  = path.compile(path);

src/core/column-list/generate.js
```diff
@@ -0,0 +1,54 @@
...
+	const props = Object.getOwnPropertyNames(source);
+	return props.reduce((acc, prop) => {
+		const key = path !== '' ?
+			`${path}.${prop}` :
+			prop;
+
+		if (deep && isObject(source[prop])) {
+			if (isArray(source[prop])) {
+				const parts = path !== '' ? path.split('.') : [];
+				acc.push({
```
klumba12: instead of acc use memo

src/core/column-list/generate.js
```diff
@@ -0,0 +1,54 @@
...
+
+export function generate(rows, deep=true, path='') {
+	if (!rows || rows.length === 0) {
+		return [];
+	}
+
+	const source = rows[0];
+	const props = Object.getOwnPropertyNames(source);
+	return props.reduce((acc, prop) => {
+		const key = path !== '' ?
```
klumba12: put expression to one line

docs/column.generation.md
```diff
@@ -1,12 +1,74 @@
...
+}
+```
+
+Column generation has two modes:
+- flat: get first-level keys of passed rows and use it for columns
+- deep: more complex mode which will handle nested objects untill doesn't find primitive type or array; it'll create getter/setter using handled path to column as well
+
+> Note that array is converted to string joined by semicolon
+
+User should set attribute generate to true for columns and generate-mode to deep|flat (deep by default) to switch on column autogeneration.
```
vkorolev: Maybe `first-level` or `shallow` instead of `flat`?

klumba12 (approved):

PR #87, opened 2017-02-02
Column types

klumba12:


src/core/services/convert.js
```diff
@@ -0,0 +1,48 @@
...
+function parseNumber(value) {
+	const number = parseFloat(value);
+	if (!isNaN(number) && isFinite(number)) {
+		return number;
+	}
+
+	return null;
+}
+
+function parseIngeger(value) {
```
vkorolev, line 41: Fix typo please:)

src/core/column/column.factory.js
```diff
@@ -8,6 +13,16 @@ export default function (type, model = null) {
 	switch (type) {
 		case 'text':
 			return new TextColumn(model || TextColumn.model());
+		case 'number':
```
vkorolev, line 16: I think this stuff should be better made as a map with function `register` and `get`.

vkorolev (approved):

PR #88, opened 2017-02-03
Default pipe configuration

vkorolev:


src/core/pipe/pipe.setup.js
```diff
@@ -0,0 +1,46 @@
...
+	if (data.pipe && data.pipe.length) {
+		service.invalidate();
+		return;
+	}
+
+	if (data.rows.length) {
+		defaults(model);
+		service.invalidate();
+	} else {
+		let off = model.dataChanged.on(e => {
```
titovmx: I think, ```off``` should be const

titovmx (approved):

src/core/pipe/pipe.build.js
```diff
@@ -0,0 +1,20 @@
...
+
+	return function run(source, changes) {
+		const context = {
+			model: model,
+			source: source,
+			valueFactory: valueFactory,
+			changes: changes
+		};
+
+		// TODO: add progress
```
vkorolev, line 14: Add progress. Maybe provide events `invalidateStart`, `invalidateEnd` or like that?

PR #94, opened 2017-02-03
Material

klumba12:


src/themes/material/templates/body.array.cell.edit.html
```diff
@@ -0,0 +1,6 @@
+<div class="editor">
+	<button type="text"
```
etokareva, line 2: button type="button"?

demo/index.js
```diff
@@ -25,15 +25,21 @@ require('./index.scss');
...
+	.indexOf('?theme=default') >= 0
+	? themeDefault
+	: themeMaterial;
+
 const dependencies = [
 	ngRoute,
 	ngAnimate,
 	ngArea,
 	ngSanitize,
 	'ngMaterial', // WTF?
```
etokareva, line 40: unnecessary comment?

src/themes/material/templates/body.array.cell.edit.html
klumba12: it's not correct editor
just some html that is not used

demo/index.js
klumba12: no
its WTF)

etokareva (approved):

titovmx (approved):

PR #95, opened 2017-02-04
Groups+Checkboxes

dmitriy-kostianetskiy:


src/core/node/node.builder.js
```diff
@@ -44,7 +45,7 @@ export default function nodeBuilder(columnMap, groupBy, valueFactory, level = 0)
 			for (let i = 0, length = keys.length; i < length; i++) {
 				const key = keys[i];
 				const group = groups[key];
-				group.node.children = build(group.rows);
+				group.node.children = build(group.rows, group.node.rows);
```
klumba12: Nice! How it works before? :)))

klumba12: For style i'd love more if pass delegate to avoid checking rowIndexes ? ...:
```javascript
build(group.rows, i => group.node.rows[i] || identity)
```
I'm not sure about perfomace

demo/pages/checkable-rows/index.html
```diff
@@ -1,21 +1,23 @@
 <q-grid columns="$ctrl.columns"
 		  rows="$ctrl.rows"
 		  selection="$ctrl.selection"
-		  selection-mode="$ctrl.selectionMode"
+		  selection-mode="check"
```
klumba12, line 4: can we change it to selection-mode="row"?
i think for displaying checkbox we'll have other option?

src/core/view/view.head.js
```diff
@@ -16,20 +16,20 @@ export default class HeadView extends View {
...
-			}
-
 			if (nodes.length) {
 				const groupColumn = columnFactory('group');
 				groupColumn.model.key = groupColumn.model.key + `[0][${columns.length}]`;
 				groupColumn.rowspan = heads.length;
 				columns.push(groupColumn);
 			}
 
+			if (model.selection().mode === 'check') {
```
klumba12: will it be shown in the 2nd column? Why not first?

demo/pages/checkable-rows/index.html
dmitriy-kostianetskiy: Let's include this fix into selection feature)

PR #96, opened 2017-02-05
Column, rows and cells selection

dmitriy-kostianetskiy:


src/core/node/node.builder.js
```diff
@@ -13,16 +13,17 @@ export default function nodeBuilder(columnMap, groupBy, valueFactory, level = 0)
 
 	const column = columnMap[groupKey];
 	const getValue = valueFactory(column);
-	return rows => {
+	return (rows, getIdentity = (i) => i) => {
```
klumba12: the name should be getRowIndex, identity means return passed arg as  it is :)
https://en.wikipedia.org/wiki/Identity_function

src/core/node/node.builder.js
```diff
@@ -44,7 +45,7 @@ export default function nodeBuilder(columnMap, groupBy, valueFactory, level = 0)
 			for (let i = 0, length = keys.length; i < length; i++) {
 				const key = keys[i];
 				const group = groups[key];
-				group.node.children = build(group.rows);
+				group.node.children = build(group.rows, i => group.node.rows[i]);
```
klumba12: small perfomance imprt: 
```javascript
const node = group.node;
const rows = node.rows;
node.children = build(group.rows, i => rows[i])
```

src/core/selection/selection.model.js
```diff
@@ -3,6 +3,7 @@ import Resource from 'core/resource/resource';
 export default class SelectionModel {
 	constructor() {
 		this.mode = 'row';
```
klumba12: can you write what modes we are supporting in comments here like this.mode = 'row'; //row | cell| column ?

src/core/selection/selection.model.js
```diff
@@ -3,6 +3,7 @@ import Resource from 'core/resource/resource';
...
+		this.checkbox = false;
```
klumba12: Get from telerik Silverlight grid:
```javascript
selectionUnit = 'row | cell | column | range | mixed';
selectionMode = 'multiple | single | extended??'
```

src/ng/components/body/body.html
```diff
@@ -1,6 +1,16 @@
...
 	<td ng-repeat="$column in $view.body.columns track by $column.model.key"
-		 ng-class="[$column.model.key, $column.model.type]"
+		 ng-class="[
+			$column.model.key, 
+			$column.model.type, 
+		 	{
+				 'selected': $selection.mode('column') && $selection.state($column) ||
+								 $selection.mode('cell') && $selection.state($cell),
+			}]"
+		 ng-click="$selection.mode('column') && $selection.toggle.execute($column) ||
```
klumba12: This is bad idea to put ng-click on body(td|tr), look at body.core.js for editCell stuff.
Need to review ng-class in next versons too...we're triing to make for td.tr as less bindings as possible

src/ng/components/selection/selection.core.js
```diff
@@ -36,6 +32,31 @@ class SelectionCore extends Directive(SELECTION_CORE_NAME, {view: `^^${VIEW_CORE
...
+		return mode === currentMode;
+	}
+
+	get checkbox() {
+		return  this.view.model.selection().checkbox;
+	}
+
+	get items() {
+		switch (this.mode()) {
+			case 'row':
```
klumba12: where is 'cell'?

klumba12: also can we move selection.core to view.selection
and in template use $view.selection.toggle...
see. core/view

klumba12 (approved):

PR #97, opened 2017-02-06
Column type for row number

klumba12:


vkorolev (approved):

src/core/column-type/column.row.indicator.js
```diff
@@ -0,0 +1,30 @@
+import ColumnView from 'core/view/view.column';
```
titovmx, line 1: what's difference between this and file below (column.row.number.js)?

PR #103, opened 2017-02-08
Ui fixes

klumba12:


src/ng/components/body/td.core.js
```diff
@@ -15,6 +15,10 @@ class TdCore extends Directive(TD_CORE_NAME, {view: `^^${VIEW_CORE_NAME}`}) {
 	}
 
 	onInit() {
+		const column = this.column;
+		this.$element[0].classList.add(column.key);
```
titovmx, line 19: Do you avoid jqLite methods specially? I mean that it can be writed as ```this.$element.addClass(column.key);```

src/ng/components/toolbar/toolbar.top.html
```diff
@@ -1,14 +1,27 @@
...
-		<!--&lt;!&ndash; filter placeholder &ndash;&gt;-->
-		<!--<div class="filter"></div>-->
-	<!--</q-grid:filter>-->
-	<q-grid:options>
-		<!-- options placeholder -->
-		<div class="options">&#8942;</div>
-	</q-grid:options>
+<div class="toolbar toolbar-top" layout="column">
+	<div layout="row">
+		<div flex="100"
```
titovmx, line 3: These attributes are ng-material specific. Probably we should keep it in theme templates.

klumba12: It's not we have layout.scss with that :)

titovmx: Yeah, I see. It keeps for default theme. What's about other custom themes? :)

src/ng/components/body/td.core.js
etokareva: classList API is very useful HTML5 feature as for me

etokareva (approved):

src/ng/directives/resizable.js
```diff
@@ -44,26 +45,28 @@ class Resizable extends Directive(RESIZABLE_NAME, {stickyCore: `^^?${STICKY_CORE
...
 				.th(sticky.source)
 				.find(th => th.classList.contains(this.th.column.key));
 		}
 
-		const width = parseInt(this.$element[0].clientWidth);
-		context.startX = e.screenX - width;
-
+		const style = window.getComputedStyle(this.$element[0], null);
+		//context.min = parseFloat(style.getPropertyValue('min-width'));
+		context.width = parseFloat(style.getPropertyValue('width'));
```
titovmx: why do you return x and width to context? it's calculating just in moment of mousedown. and then const for drag

src/ng/components/body/td.core.js
klumba12: Yep, to port easy

src/ng/directives/resizable.js
klumba12: I know - just more readable as for me when do width - offsetX;
also getComptedStyle solve problem with jumping :)

src/ng/components/toolbar/toolbar.top.html
klumba12: Yep we'll have a non-theme css for the core grid itself iguess?

PR #112, opened 2017-02-11
GRID_PREFIX add

etokareva:


src/core/highlight/highlight.view.js
```diff
@@ -3,6 +3,7 @@ import Command from 'core/infrastructure/command';
 import * as columnService from 'core/column/column.service';
```
klumba12, line 3: GOOD POINT =)
Maybe defenitions we need to put to the core? or not?

etokareva: @klumba12 moved definitions to core, or maybe better to core/infrastructure?

klumba12 (approved):

PR #128, opened 2017-02-15
Selection navigation

etokareva:


src/core/selection/selection.view.js
```diff
@@ -2,43 +2,243 @@ import View from '../view/view';
 import Command from 'core/infrastructure/command';
 import behaviorFactory from './selection.factory';
```
vkorolev, line 3: This comment is not about this PR. Maybe we should have `strategyFactory` instead of `behaviorFactory`?
Strategy is a common abstraction to encapsulate behavior

src/core/selection/selection.view.js
```diff
@@ -2,43 +2,243 @@ import View from '../view/view';
...
+	highlight(item) {
+		item.classList.add(`${GRID_PREFIX}-selected`);
+	}
+
+	blur(item) {
+		item.classList.remove(`${GRID_PREFIX}-selected`);
+	}
+
+	get commands() {
+		const model = this.model;
```
vkorolev: Consider to move commands to separate file. Kind of `commandsFactory = model => ...`

src/core/selection/selection.view.js
```diff
@@ -2,43 +2,243 @@ import View from '../view/view';
...
+					//highlight newValue columns
 				}
-
-				model.selection({items: this.behavior.view}, {source: 'toggle'});
 			}
 		});
+	}
 
-		model.viewChanged.watch(() => this.behavior = behaviorFactory(model));
+	highlightRow(item) {
```
vkorolev: Consider to move all highlight/blur stuff to a separate file.

src/core/selection/selection.view.js
etokareva: Like commands.selection.js and commands.navigation.js?

vkorolev: Yes, but factory-function would be better than class.

The same for navigation/navigation .js

src/core/selection/selection.view.js
klumba12: As strategy is a part of behavior algorithms it can be renamed, but from readable context i like behavior more here

klumba12: Selection.strategy vs Selection.behavior

src/core/selection/selection.view.js
```diff
@@ -2,43 +2,243 @@ import View from '../view/view';
...
+		model.selectionChanged.watch(e => {
+			if (e.tag.source !== 'toggle'
+				&& e.changes.hasOwnProperty('items')) {
+				this.behavior.select(e.state.items, true);
+			}
 
-				if (isUndefined(state)) {
-					state = this.behavior.state(item);
-					this.behavior.select(item, state === null || !state);
+			if (e.changes.hasOwnProperty('unit')) {
```
klumba12: can be deleted

src/core/selection/selection.view.js
```diff
@@ -2,43 +2,243 @@ import View from '../view/view';
...
-					item = model.view().rows;
-				}
+		model.viewChanged.watch(e => {
+			this.behavior = behaviorFactory(model);
+			if (e.changes.hasOwnProperty('rows')) {
+				e.state.rows.forEach((item) => this.blurRow(item));
+			}
+		});
+		model.selectionChanged.watch(e => {
+			if (e.tag.source !== 'toggle'
```
klumba12: put that under one if e.changes.hasOwnProperty('items')

klumba12 (approved):

PR #130, opened 2017-02-16
Select all for column chooser and column filter

klumba12:


src/ng/plugins/column-chooser/column.chooser.js
```diff
@@ -123,18 +147,30 @@ class ColumnChooser extends Plugin {
...
-		orderFromViewToData(this._columns, model.view().columns[0]);
+		orderFromViewToData(this.columns, model.view().columns[0] || []);
 	}
 
 	state(column) {
 		return column.isVisible !== false;
 	}
 
+	stateAll() {
+		return this.columns.every(this.state.bind(this));
```
vkorolev, line 163: Can we use the whole power of ES6 here please (lambdas i mean)?:)

vkorolev (approved):

PR #145, opened 2017-02-21
Duplicate selection column issue fixed

dmitriy-kostianetskiy:


src/core/services/convert.js
```diff
@@ -62,6 +64,10 @@ export function isPrimitive(type) {
 	}
 }
 
+function parseBool(value){
+	return value === 'true';
```
vkorolev: In this case a string like `'some-string'` will be parsed as `false`. 
I'm not sure it's consistent behavior.

I think something like this should be:
```javascript
   return value === 'true' 
      ? true 
      : value === 'false' 
         ? false 
         : null;
```

dmitriy-kostianetskiy: Cool! Updated)

vkorolev (approved):

PR #149, opened 2017-03-01
Pin initial thoughts

etokareva:


src/core/pin/pin.model.js
```diff
@@ -0,0 +1,9 @@
+import Resource from '../resource/resource';
+
+export default class PinModel {
+	constructor() {
+		this.resource = new Resource();
+		this.by = [];
```
klumba12: If we have pin property in column why we need that?

src/core/pin/pin.view.js
```diff
@@ -0,0 +1,33 @@
+import View from 'core/view/view';
+import * as pinService from 'core/pin/pin.service';
+import Command from 'core/infrastructure/command';
+import * as columnService from 'core/column/column.service';
+
+export default class PinView extends View {
+	constructor(model) {
+		super(model);
```
klumba12, line 8: This view should expose 2 properties left and right that are lists of columns

src/ng/components/grid/grid.js
```diff
@@ -117,6 +117,8 @@ export default {
 		pivotBy: '<',
 		sortBy: '<',
 		sortMode: '@',
+		pinBy: '<',
+		pinMode: '@',
```
klumba12: This is not required pin is property of column, mode doesnt make value

PR #152, opened 2017-03-03
Header titles disappearing bug

titovmx:
- fixed title disappearing - parseDate method set date type to ```null``` value mistakenly
- fixed order of new columns from column-list

### DIFF
--- a/src/core/services/convert.js
+++ b/src/core/services/convert.js
@@ -77,6 +77,10 @@ function parseText(value) {
 }
 
 function parseDate(value) {
+	if (value === null) {
+		return null;
+	}
+
 	const date = new Date(value);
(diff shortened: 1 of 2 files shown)


vkorolev (approved):

src/core/services/convert.js
```diff
@@ -77,6 +77,10 @@ function parseText(value) {
 }
 
 function parseDate(value) {
+	if (value === null) {
```
klumba12, line 80: check another parse functino

klumba12 (approved):

src/core/services/convert.js
titovmx: I'm going to cover it by tests. I'll create another pr for this.

PR #169, opened 2017-03-14
Range selection improments

dmitriy-kostianetskiy:


klumba12 (approved):

demo/pages/selection-model/index.js
```diff
@@ -1,22 +1,34 @@
...
+
 		$log.log(`qgrid.demo: selection changed ${e.state.items.length} on ${e.state.unit} unit and ${e.state.mode} mode`);
 	};
 
-	$http.get('data/people/100.json')
-		.then(function (response) {
-			ctrl.rows = response.data;
-			ctrl.selection = response.data.slice(0, 4).map((row) => {
+	this.model.dataChanged.watch((e) => {
+		if (e && e.changes.hasOwnProperty('columns') && e.state.columns.length > 0 && e.state.rows.length > 0) {
```
klumba12, line 18: don't need to check on e right now is always not null

src/core/selection/state/multiple.selection.state.js
```diff
@@ -3,27 +3,28 @@ import SelectionState from './selection.state';
...
 	}
 
 	selectCore(item, state) {
 		if (state) {
-			this.items.add(item);
+			this.items.set(this.key(item), item);
 		}
 		else {
-			this.items.delete(item);
+			this.items.delete(this.key(item));
```
klumba12, line 15: remove is more common name, also delete is system operation

src/ng/components/body/body.core.js
```diff
@@ -76,48 +80,42 @@ class BodyCore extends Directive(BODY_CORE_NAME, {view: `^^${VIEW_CORE_NAME}`})
...
+		const endCell = pathFinder.cell(e.path);
+		
+		if (startCell && endCell) {
+			this.view.selection.behavior.selectRange(startCell, endCell);
+						
+			this.view.model.navigation({
+				active: {
+					cell: endCell
+				},
+				column: endCell.$element[0].cellIndex,
```
klumba12, line 102: why we can't use endCell.rowIndex|columnIndex?

demo/pages/selection-model/index.js
klumba12: Also use e.hasChanges method

src/core/selection/state/multiple.selection.state.js
dmitriy-kostianetskiy: That's built-in ES2015 Map. I cannot delete this method

PR #177, opened 2017-03-20
Table

etokareva:
replace of markup (except highlight.view)


src/ng/services/table.js
```diff
@@ -1,3 +1,25 @@
+import {noop} from 'core/services/utility';
+class Dom {
+	constructor(element) {
+		this.element = element;
+	}
+
+	column(index) {
+		let cells = [];
```
klumba12: Check index ranges

src/ng/services/table.js
```diff
@@ -19,4 +41,13 @@ export default class Table {
 
 		return null;
 	}
+	get body(){
+		return new Dom(this.markup.body) || noop;
```
klumba12: shouldnt be noop should be something like this.markup.body ? new Dom(this.markup.body) : Dom.empty

src/ng/services/table.js
```diff
@@ -1,3 +1,25 @@
...
+		this.element = element;
+	}
+
+	column(index) {
+		let cells = [];
+		const rows = this.element.rows;
+		for (let i = 0; i++; i < rows.length) {
+			cells.push(rows[i].cells[index]);
+		}
+		return {
```
klumba12: should be in chain style column(index).cells()

src/ng/services/table.js
```diff
@@ -1,3 +1,25 @@
...
+			cells: cells
+		}
+	}
+
+	row(index) {
+		return this.element.rows[index];
```
klumba12: should be row(index).cells()

src/ng/services/table.js
etokareva: Here, in table.js?

src/ng/services/table.js
```diff
@@ -1,3 +1,40 @@
+let instance = {};
```
klumba12: what will happen when empty.column(0) will be called?

src/ng/services/table.js
```diff
@@ -1,3 +1,40 @@
...
+			if (index >= 0 && index < rows.length) {
+				for (let i = 0; i++; i < rows.length) {
+					cells.push(rows[i].cells[index]);
+				}
+			}
+			return cells;
+		}
+
+		return {
+			cells: cells()
```
klumba12: still is not lazy

src/ng/services/table.js
```diff
@@ -1,3 +1,58 @@
+class Dom {
+	constructor(element) {
+		this.element = element;
+	}
+
+	static get empty() {
+		return new Dom(null);
```
klumba12: The better idea will be the base class with empty methods , check on nulls is always worse

src/ng/services/table.js
```diff
@@ -1,3 +1,58 @@
...
+	}
+
+	column(index) {
+		const element = this.element;
+		if (!element) {
+			return {cells: []};
+		}
+		const rows = element.rows;
+		const cellsCount = rows[0].cells.length;
+		let obj = {};
```
klumba12: Can we have separate class for that? Row?

src/ng/services/table.js
```diff
@@ -1,3 +1,58 @@
+class Dom {
```
klumba12, line 1: Rename dom to TableDom, we need RowDom and ColumnDom also

src/ng/services/table.js
```diff
@@ -1,3 +1,80 @@
...
+					configurable: true,
+					enumerable: false,
+					value: value,
+					writable: true
+				});
+			}
+		});
+	}
+}
+class Empty {
```
klumba12: Rename it to Dom =)
```javascript
TableDom extends Dom
```

src/ng/services/table.js
```diff
@@ -1,3 +1,80 @@
...
+		const cellsCount = rows[0].cells.length;
+		if (row >= 0 && row < rows.length && column >= 0 && column < cellsCount) {
+			return rows[row].cells[column];
+		}
+	}
+}
+
 export default class Table {
 	constructor(markup) {
 		this.markup = markup;
```
klumba12: lets remove cellAt method from the table

src/ng/services/table.js
```diff
@@ -1,3 +1,80 @@
...
+	static get empty() {
+		return new Empty();
+	}
+
+	column(index) {
+		const element = this.element;
+		const rows = element.rows;
+		const cellsCount = rows[0].cells.length;
+		let obj = {};
+		new lazyProperty(obj, "cells", () => {
```
klumba12: why not just a create?
```javascript
return new ColumnDom(...
```

src/ng/services/table.js
```diff
@@ -1,3 +1,80 @@
...
+					result.push(rows[i].cells[index]);
+				}
+			}
+			return result;
+		});
+		return obj;
+	}
+
+	row(index) {
+		const element = this.element;
```
klumba12: this should return RowDom and one of lazy prop should be cells, we're going to encapsulate dom as much as possible, let for now only cell element be available

src/ng/services/table.js
etokareva: I wanted to do it, but I think we need it in navigation(to get cell from keyboard)

src/ng/services/table.js
```diff
@@ -1,3 +1,96 @@
...
+		}
+	}
+}
+class ColumnDom extends Dom {
+	constructor(element, index) {
+		super(element);
+		const rows = element.rows;
+		const cellsCount = rows[0].cells.length;
+		let obj = {};
+		new Create(obj, "cells", () => {
```
klumba12: Why not just use es6 syntax 
```javascript
get cells(){
...
}
```

src/ng/services/table.js
klumba12: table.body.cell(i, j)?

src/ng/services/table.js
```diff
@@ -1,22 +1,147 @@
...
+		this.index = index;
+	}
+
+	get cells() {
+		const rows = this.element.rows;
+		const index = this.index;
+		if (index >= 0 && index < rows.length) {
+			return rows[index].cells
+		}
+	}
```
klumba12: Should return empty array if not in range

src/ng/services/table.js
```diff
@@ -1,22 +1,147 @@
...
+		super(element);
+		this.element = element;
+		this.index = index;
+	}
+
+	get cells() {
+		const rows = this.element.rows;
+		const index = this.index;
+		if (index >= 0 && index < rows.length) {
+			return rows[index].cells
```
klumba12: this shoud return array of DomCell

src/ng/services/table.js
```diff
@@ -1,22 +1,147 @@
...
+class CellDom extends Dom {
+	constructor(element, row, column) {
+		super(element);
+		this.element = element;
+		this.row = row;
+		this.column = column;
+	}
+
+	get view() {
+		const element = this.element;
```
klumba12: why not return just element?

src/ng/services/table.js
```diff
@@ -1,22 +1,147 @@
...
+	row(index) {
+		return new RowDom(this.element, index);
+	}
+
+	cell(row, column) {
+		return new CellDom(this.element, row, column);
+	}
+}
+class CellDom extends Dom {
+	constructor(element, row, column) {
```
klumba12: element should be cell element

src/ng/services/table.js
klumba12: and we decided to not give access to dom elements theirselfs

src/ng/services/table.js
```diff
@@ -1,22 +1,143 @@
...
+		const rows = this.element.rows;
+		if (index >= 0 && index < rows.length) {
+			return new RowDom(rows[index]);
+		}
+		return null;
+	}
+
+	cell(row, column) {
+		const rows = this.element.rows;
+		row = row < 0 ? 0 : row;
```
klumba12: this not correct behavior, to change index, if user pass -1 should return null

klumba12: Also, Kate, can you add tests for that file, please? Cause that will be essentual part of the grid

src/ng/services/table.js
```diff
@@ -1,22 +1,143 @@
...
+}
+
+class RowDom extends Dom {
+	constructor(element) {
+		super(element);
+		this.element = element;
+	}
+
+	get view() {
+		return this.element;
```
klumba12: we won't propogate element to the user

src/ng/services/table.js
```diff
@@ -1,22 +1,143 @@
...
+	}
+
+	get cells() {
+		const row = this.element;
```
klumba12: I think we shouldn't make any property as getter, cause it will force user to make lake that
```javascript
if(table.body.row(0).cells.length){
table.body.row(0).cells.forEach
}
```
User won't see the perfomance impact of using `cells` 2 times

src/ng/services/table.js
```diff
@@ -1,22 +1,143 @@
...
+		const cellsCount = row.cells.length;
+		let result = [];
```
klumba12: should be const result

src/ng/services/table.js
```diff
@@ -1,22 +1,143 @@
...
+		return new Dom();
+	}
+
+	get view() {
+		return this.element;
+	}
+
+	column(index) {
+		return new ColumnDom(this.element, index);
+	}
```
klumba12: Column and row should have the same behavior on checking index. For row if index not correct - null is return, for column ColumnDom.

src/ng/services/table.js
etokareva: And if it is not a getter - than user won't be able to do such thing?

klumba12: why?
```javascript
const cells = table.body.row(0).cells();
if(cells.length...
```
it will just force user to persist cells to own variable

src/ng/services/table.js
etokareva: If we don't do it - then I don't know how incapsulate things such as body.rows[index].clientHeight
Or can I use element property?

src/ng/services/table.js
etokareva: Why he couldn't create own var like this :
`const cells = table.body.row(0).cells`
?

src/ng/services/table.js
etokareva: Okay

src/core/highlight/highlight.view.js
```diff
@@ -117,67 +118,69 @@ export default class HighlightView extends View {
...
-					row.cells[index].classList.remove(`${GRID_PREFIX}-${cls}`);
-					if (index > 0) {
-						row.cells[index - 1].classList.remove(`${GRID_PREFIX}-${cls}-prev`);
-					}
-
-					if (index < head.rows.length - 1) {
-						row.cells[index + 1].classList.remove(`${GRID_PREFIX}-${cls}-next`);
-					}
+			const head = table.head;
+			if (markup.head.rows.length) {
```
klumba12, line 164: if(markup.head.rows.length) - do we need that?

src/core/highlight/highlight.view.js
```diff
@@ -117,67 +118,69 @@ export default class HighlightView extends View {
...
+		if (markup.body.rows.length) {
+			const cells = table.body.column(index).cells();
+			cells.forEach((cell) => cell.addClass(`${GRID_PREFIX}-${cls}`));
 		}
-
-		const foot = this.markup.foot;
-		for (let row of foot.rows) {
-			row.cells[index].classList.add(`${GRID_PREFIX}-${cls}`);
+		if (markup.foot.rows.length) {
+			const cells = table.foot.column(index).cells();
```
klumba12, line 147: looks much better!

src/ng/services/table.js
```diff
@@ -1,22 +1,134 @@
...
+		super();
+		this.element = element;
+		this.index = index;
+	}
+
+	cells() {
+		const index = this.index;
+		const rows = this.element.rows;
+		const result = [];
+		for (let i = 0; i < rows.length; i++) {
```
klumba12: lets use common style
```javascript
for(let i =0, length = rows.length; i < length; i++)...
```

klumba12: Use it in each for loop please)

src/ng/services/table.spec.js
```diff
@@ -0,0 +1,92 @@
...
+		table: htmlToElement(table),
+		body: htmlToElement(body),
+		head: htmlToElement(head)
+	};
+
+	const markupEmpty = {};
+	const tableArr = [new Table(markup), new Table(markupEmpty)];
+	const tableProps = ['body', 'head', 'foot'];
+
+	describe('table should have properties and they should be objects', () => {
```
klumba12: Actually there is no much sence to check if poperty is exists, all this properties should be covered by unit tests

src/ng/services/table.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+const myWindow = window.open("", "MsgWindow", "width=100,height=50");
+function htmlToElement(html) {
+	const template = myWindow.document.createElement('template');
+	template.innerHTML = html;
+	return template.content.firstChild;
+}
+
+describe('Table', () => {
+	const table = '<table tabindex=\"0\"><thead><tr><th><label aria-label=\"Last Name\">' +
```
klumba12: I'm not big fun of testing with dom elements... At least why we need all this attributes? tabindex=0 arai-label? Please remove it

src/ng/services/table.spec.js
```diff
@@ -0,0 +1,92 @@
...
+				});
+			}
+		});
+	});
+	describe('empty markup call to properties check', () => {
+		const table = tableArr[1];
+		it('column', () => {
+			const column = table.body.column();
+			const cells = column.cells();
+			expect(column).to.be.an.instanceOf(Object).and.to.have.property('cells');
```
klumba12, line 34: no need to check property cells =)
Especially when before the assert there is column.cells() method

src/ng/services/table.spec.js
```diff
@@ -0,0 +1,92 @@
...
+		});
+	});
+	describe('normal markup call to properties check', () => {
+		const table = tableArr[0];
+		it('column', () => {
+			const body = table.body;
+			const column = body.column(0);
+			const cells = column.cells();
+			expect(column).to.be.an.instanceOf(Object).and.to.have.property('cells');
+			expect(cells).to.be.an.instanceOf(Array).with.lengthOf(4);
```
klumba12: this is a better test

src/ng/services/table.spec.js
```diff
@@ -0,0 +1,92 @@
+import Table from './table';
```
klumba12, line 1: Tests should check logic not interface.
Example of usual test
```javascript
const cell34 = '<td....' // better to have virtual td from my opinion
....
table.body.row(3).cell(4) should be equal cell34
```

src/core/highlight/highlight.view.js
etokareva: Unfortunately, I face the TypeError: Cannot read property 'cells' of undefined if we haven't this condition on th hover event

src/ng/services/table.spec.js
etokareva: By the way, should we have functionality like in this example? To access cell by index inside row and column classes?

src/ng/services/table.spec.js
etokareva: Attributes can be deleted. It's just not fully cleaned up table, I think :-)

klumba12 (approved):

PR #188, opened 2017-03-23
Style api

klumba12:


src/core/style/style.model.js
```diff
@@ -0,0 +1,10 @@
+import Resource from '../resource/resource';
+import {noop} from '../services/utility';
+
+export default class SortModel {
```
vkorolev: Minor one: StyleModel instead of SortModel

titovmx (approved):

demo/pages/style/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+	$http.get('data/people/100.json')
+		.then(response => this.rows = response.data);
+
+	const randomColor = () => `${Math.floor(Math.random() * 16777215).toString(16)}`;
+	this.femaleColor = randomColor();
+	this.maleColor = randomColor();
+
+	this.model.style({
+		row: (row, dom) => {
```
vkorolev: Maybe row, element ? dom is less specific as for me

vkorolev (approved):

demo/pages/style/index.js
klumba12: Its nor dom nor element just instance with one method class, and this is by design. Any ideas? Maybe add this method to context?

PR #201, opened 2017-03-30
Keyboard nav in edit mode

etokareva:
Also I wanted to change shortcut for Enter? or Tab? to smth like model.navigation().active.cell? model.navigation().active.cell.column.editorSubmitRequest : 'tab', to make it depend on editor-submit-request, but I'm not sure in this. Maybe we should redesign commands...


src/core/column-type/number.column.js
```diff
@@ -19,6 +19,7 @@ TemplatePath.register('number-cell-edit', (template, column) => {
 class NumberColumnModel extends DataColumnModel{
 	constructor(){
 		super('number');
+		this.editorSubmitShortcut = 'tab';
```
klumba12: I think we will put to the edit model, and remove from column. It doesn't look natural here. Can you put it to
model.edit().commitShortcuts = {
'$default': 'tab|enter',
'date': 'enter',
...
};

something like that? or maybe in navigation model?

src/core/edit/edit.cell.view.js
```diff
@@ -34,20 +34,19 @@ export default class EditCellView {
 				canExecute: cell => {
 					cell = cell || model.navigation().active.cell;
 					if (cell && this.mode !== 'edit' && model.edit().mode === 'cell' && cell) {
-						return cell.column.canEdit && model.edit().enter.canExecute(this.contextFactory(cell));
+						return cell.column.canEdit && model.edit().enter.canExecute(this.contextFactory(cell)) && model.edit().editMode === 'view';
 					}
 
-					return false;
+					return model.edit().editMode === 'view';
```
klumba12: why?

etokareva: You're right. Here we don't need it.

src/core/edit/edit.cell.view.js
```diff
@@ -161,6 +160,21 @@ export default class EditCellView {
 		this._value = value;
 	}
 
+	get commitShortcut() {
+		const commitShortcuts = {
```
klumba12, line 164: Kate, can you put to to the edit.model?

src/ng/components/column/column.js
```diff
@@ -31,6 +31,9 @@ class Column extends Component {
 				this.key = '$default';
 			}
 		}
+		if ($attrs.hasOwnProperty('editorSubmitShortcut')) {
```
klumba12, line 34: lets remove that for now

klumba12 (approved):

PR #202, opened 2017-03-30
Column order fixes

klumba12:


src/core/navigation/navigation.js
```diff
@@ -17,157 +15,230 @@ export default class Navigation {
...
-					if (navigationState.row == -1 && navigationState.column == -1) {
-						model.navigation({
-							column: 0,
-							row: 0
-						});
-					} else {
-						model.navigation({row: navigationState.row + 1});
-					}
-				}
+				canExecute: () => canExecute() && this.nextRow >= 0,
```
etokareva, line 154: Why we use canExecute inside of canExecute?

klumba12: this is common canExecute logic for all navigation commands, What name do you prefer?

etokareva: Just don't fully understand how the code works

klumba12: just look on canExecute method that is defined above

titovmx (approved):

PR #207, opened 2017-04-03
improvements

etokareva:


src/core/edit/edit.model.js
```diff
@@ -11,4 +11,19 @@ export default class EditModel {
 		this.cancel = new Command();
 		this.reset = new Command();
 	}
+
+	get commitShortcut() {
```
klumba12: model should have only commitShortcuts option

klumba12: all other should be in controller

src/core/edit/edit.model.js
```diff
@@ -10,5 +10,11 @@ export default class EditModel {
 		this.commit = new Command();
 		this.cancel = new Command();
 		this.reset = new Command();
+		this.commitShortcuts = {
+			'$default': 'tab|enter',
```
klumba12: please for the date add ctrl+s

klumba12: for the text tab is ok too, also tab should go to the next cell from edit mode does it work?

etokareva: With tabs and shift tabs everything is ok. ctrl+s causes saving of the html also as commiting of the text/number/password templates. But date template seems to not work with ctrl+s

vkorolev: `ctrl+s` is a common system shortcut, so we I don't think we need to intercept this

klumba12 (approved):

PR #213, opened 2017-04-05
Bug fixes scroll and selection

klumba12:


src/core/selection/selection.view.js
```diff
@@ -194,11 +195,21 @@ export default class SelectionView extends View {
 	}
 
 	select(items) {
-		if (this.selection.mode === 'range') {
-			this.selectionState.clear();
-			this.selectionState.toggle(items, true);
-		} else {
-			this.selectionState.toggle(items);
+		if (arguments.length && !isUndefined(items)) {
```
vkorolev, line 198: I think `arguments.length` is unnecessary

klumba12: the problem is that isUndefined shouldn't be here but as we use lambdas for now in commands i just can't pass ...argumnets, so this is notification to refactor in the future

vkorolev (approved):

PR #215, opened 2017-04-06
Feature/advanced column types

titovmx:
- Time
- Url
- Email
- File (single upload)
- Image


demo/pages/edit-templates/index.js
```diff
@@ -83,6 +83,43 @@ export default function Controller($http) {
...
+			label: (item, label) => isUndef(label)
+				? item.webPageLabel || `${item.name.last} ${item.name.first}`
+				: item.webPageLabel = label
+		},
+		{
+			key: 'attachment',
+			title: 'Attachment',
+			type: 'file',
+			value: (item, value) => isUndef(value) ? item.attachment : item.attachment = value,
+			onUpload: (e) => {
```
klumba12: do we need  ut onUpload to the editor-options?

src/core/column-type/file.column.js
```diff
@@ -0,0 +1,41 @@
...
+	};
+});
+
+class FileColumnModel extends DataColumnModel {
+	constructor() {
+		super('file');
+
+		this.canEdit = false;
+		this.onUpload = noop;
+		this.canUpload = yes;
```
klumba12: this should be in the editor options too

src/core/column-type/file.column.js
```diff
@@ -0,0 +1,41 @@
...
+
+		this.hasPreview =
```
klumba12: this too

src/ng/directives/file.upload.js
```diff
@@ -0,0 +1,75 @@
...
+		if (files.length > 1) {
+			throw new AppError('file.upload', `Multiple upload isn't able`);
+		}
+		
+		this.file = files[0] || null;
+		if (this.file) {
+			this.reader.readAsDataURL(this.file);
+		}
+		
+		this.$scope.$evalAsync(() => this.onUpload({
```
klumba12: it should be applyAsync, event doesn't trigger digest

src/core/edit/edit.cell.view.js
```diff
@@ -160,6 +170,14 @@ export default class EditCellView {
 		this._value = value;
 	}
 
+	get label() {
+		return this._label;
+	}
+
+	set label(label) {
```
klumba12, line 180: need to think about that

demo/pages/edit-templates/index.js
vkorolev: Depends on use-case. If items are uploaded separately, it makes sense.
But i'd rather have `upload` instead of it, because only user knows how to upload file

src/ng/directives/file.upload.js
vkorolev: Imo it should be upload which returns promise, so it is triggered like this:
```javascript
this
    .upload(this.file)
    .then(() => this.$scope.$evalAsync())
```

demo/pages/edit-templates/index.js
```diff
@@ -1,13 +1,38 @@
...
+			if(e.column.key === 'attachment' || e.column.key === 'avatar'){
+				$timeout(() => {
+					const filename = e.newLabel;
+					$mdToast.show(
+						$mdToast.simple()
+							.textContent(`File ${filename} loaded`)
+							.position('top right')
+							.hideDelay(2000)
+					);
+					e.column.value(e.row, `https://fake.data.server.com/attachment/${encodeURI(filename)}`);
```
klumba12, line 19: use valueFactory

src/core/column-type/file.column.js
```diff
@@ -0,0 +1,41 @@
...
+		model: 'edit',
+		resource: column.key
+	};
+});
+
+class FileColumnModel extends DataColumnModel {
+	constructor() {
+		super('file');
+
+		this.canEdit = false;
```
klumba12, line 25: will refactor that to editTrigger

klumba12 (approved):

PR #216, opened 2017-04-10
Markup refactoring for navigation

etokareva:


src/ng/services/dom/box.js
```diff
@@ -159,4 +159,8 @@ export default class Box extends BoxCore {
 
 		this.element.scrollTop = value;
 	}
+
+	get bounds() {
+		return this.element.getBoundingClientRect();
```
klumba12: the name should be "rect" at least and make it as function please or make scrollTop scrollLeft as getters

klumba12 (approved):

PR #217, opened 2017-04-10
Release script

vkorolev:
* Add ShellJS dependency
* Add release.js script
* Add `npm run release` script


scripts/release.js
```diff
@@ -0,0 +1,46 @@
...
+const fs = require('fs');
+
+const RELEASE_BRANCH = '1.0.x';
+const cwd = path.resolve('.');
+const pkgPath = path.join(cwd, 'package.json');
+const bowerPath = path.join(cwd, 'bower.json');
+
+function main() {
+	shell.exec(`git checkout ${RELEASE_BRANCH}`);
+	shell.exec(`git stash`);
```
titovmx: Maybe there's better place ```stash``` before ```checkout```? And make it sense to do ```git stash pop``` at the end?

klumba12 (approved):

scripts/release.js
klumba12: why we need stash here?

vkorolev: Stash before checkout is a good point. To use `stash pop` we need more "smart" script to persist current branch, and track if any changes were stashed

PR #218, opened 2017-04-10
Place value and label services to core

titovmx:


src/core/body/body.view.js
```diff
@@ -83,13 +85,18 @@ export default class BodyView extends View {
 		};
 	}
 
+	labelFactory(column) {
+		const getLabel = thus._labelFactory(column);
```
vkorolev: Type: `this` instead of `thus` i guess

titovmx: Oh, sure. Thank you!

klumba12 (approved):

PR #219, opened 2017-04-11
Empty

etokareva:


src/ng/services/dom/box.js
```diff
@@ -28,7 +28,7 @@ class BoxCore {
 	}
 
 	cell() {
-		return null;
```
vkorolev: It doesn't look useful. Do we use it anywhere?

src/ng/services/dom/element.js
```diff
@@ -1,9 +1,42 @@
...
+		return 0;
+	}
+}
 export default class Element {
 	constructor(element) {
 		this.element = element;
 	}
 
+	static get empty() {
+		return new ElementCore();
```
vkorolev, line 33: Consider to make it singleton and avoid instantiation on every call

src/ng/services/dom/table.spec.js
```diff
@@ -43,7 +43,7 @@ describe('Table', () => {
 		});
 		it('cell', () => {
 			tableProps.forEach((prop) => {
-				expect(table[prop].cell(0, 0)).to.be.null;
+				expect(table[prop].cell(0, 0)).to.be.an.instanceOf(Object);
```
vkorolev, line 46: `to.equal(Cell.empty)` maybe?

src/ng/services/dom/element.js
etokareva: Ok, it was attemt to create singleton))) I'll try again.

src/ng/services/dom/cell.js
```diff
@@ -2,16 +2,16 @@ import Element from './element';
 import AppError from 'core/infrastructure/error';
 
 export default class Cell extends Element {
+	static get empty() {
+		return Element.empty;
+	}
```
klumba12, line 7: why element.empty? does element contains all cell properties?

src/ng/services/dom/element.js
```diff
@@ -1,9 +1,42 @@
...
+	constructor() {
+	}
+
+	get model() {
+		return null;
+	}
+
+	rect() {
+		return {};
+	}
```
klumba12, line 11: object should have top,left,width,height equal to zero?

etokareva:
Closed as it was pulled to markup-refactoring branch

PR #220, opened 2017-04-11
Markup refactoring for highlight

etokareva:


src/core/highlight/highlight.view.js
```diff
@@ -150,32 +148,31 @@ export default class HighlightView extends View {
...
 				}
-				if (index < markup.head.rows.length - 1) {
+				if (index < table.head.rowCount() - 1) {
 					const cells = head.column(index + 1).cells();
 					cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}-next`));
 				}
 			}
 
-			if (markup.body.rows.length) {
+			if (table.body.rowCount()) {
```
klumba12: why you need that check if cells will just return empty array?

klumba12: So the aim of the table.js to get rid of all if statements where possible too

src/core/highlight/cell.selector.js
```diff
@@ -1,15 +1,15 @@
...
+export default (model, table) => {
 	function getRows(items) {
 		const result = [];
 		const rows = model.view().rows;
 		
 		for (let item of items) {
 			const index = rows.indexOf(item);
-			if (index > -1 && markup.body.rows[index]) {
-				for (let cell of markup.body.rows[index].cells) {
+			if (index > -1 && table.body.row(index)) {
```
klumba12: should return empty row so no need to check indicies in the future i guess :)

src/ng/services/dom/cell.js
```diff
@@ -1,7 +1,44 @@
-import Element from './element';
 import AppError from 'core/infrastructure/error';
+import Element from './element';
+
+class CellCore {
+	get model() {
+		return null;
+	}
+
+	rect() {
```
klumba12: This is the same like in element, can we extend all core from elementcore?

src/core/highlight/highlight.view.js
```diff
@@ -116,69 +115,54 @@ export default class HighlightView extends View {
...
 		}
-		if (markup.body.rows.length) {
-			const cells = table.body.column(index).cells();
-			cells.forEach((cell) => cell.addClass(`${GRID_PREFIX}-${cls}`));
-		}
-		if (markup.foot.rows.length) {
-			const cells = table.foot.column(index).cells();
-			cells.forEach((cell) => cell.addClass(`${GRID_PREFIX}-${cls}`));
+
+		if (index < table.head.rowCount() - 1) {
```
titovmx: Is it ```index``` of column? Why here is ```rowCount()```, but not ```columnCount()```?
Besides, minor thing, you had already put ```table.head``` to variable ```head``` and can use it

src/core/highlight/highlight.view.js
```diff
@@ -116,69 +115,54 @@ export default class HighlightView extends View {
...
+			const headCells = head.column(index).cells();
+			headCells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}`));
+			if (index > 0) {
+				const cells = head.column(index - 1).cells();
+				cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}-prev`));
 			}
-			if (markup.foot.rows.length) {
-				const cells = table.foot.column(index).cells();
-				cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}`));
+			if (index < table.head.rowCount() - 1) {
```
titovmx: The same question about this row

src/ng/services/dom/table.spec.js
```diff
@@ -32,18 +35,18 @@ describe('Table', () => {
 		it('column', () => {
 			const column = table.body.column();
 			const cells = column.cells();
-			expect(column).to.be.an.instanceOf(Object).and.to.have.property('cells');
+			expect(column).to.deep.equal(Column.empty);
```
vkorolev: I think, `deep` is unnecessary since `*.empty` is singleton now

src/core/highlight/highlight.view.js
```diff
@@ -116,69 +115,54 @@ export default class HighlightView extends View {
...
-					cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}-next`));
-				}
-			}
-
-			if (markup.body.rows.length) {
-				const cells = table.body.column(index).cells();
-				cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}`));
+			const headCells = head.column(index).cells();
+			headCells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}`));
+			if (index > 0) {
```
klumba12: Why we need checks on indicies here?

src/core/highlight/highlight.view.js
```diff
@@ -116,69 +115,54 @@ export default class HighlightView extends View {
...
+				const cells = head.column(index - 1).cells();
+				cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}-prev`));
 			}
-			if (markup.foot.rows.length) {
-				const cells = table.foot.column(index).cells();
-				cells.forEach((cell) => cell.removeClass(`${GRID_PREFIX}-${cls}`));
+			if (index < head.rowCount() - 1) {
```
klumba12: And here

src/ng/services/dom/cell.js
```diff
@@ -1,7 +1,20 @@
-import Element from './element';
 import AppError from 'core/infrastructure/error';
+import Element from './element';
+import ElementCore from './elementCore';
+
```
klumba12: Core classes should be in the same files just export them

src/core/highlight/highlight.view.js
etokareva: I've looked at the annotations and don't know why had you added them...

src/core/highlight/highlight.view.js
etokareva: And there

src/ng/services/dom/cell.js
etokareva: ok

klumba12 (approved):

PR #222, opened 2017-04-14
autocomplete editor

titovmx:
- add autocomplete editor
- refactor initClick directive to raise event
- add example to editors demo page


src/core/infrastructure/fetch.js
```diff
@@ -26,7 +26,8 @@ export default class Fetch {
 					reject: rejectBusy
 				};
 
-				const result = select(...arguments, deferred);
+				const args = Array.from(arguments).slice(1) || [];
```
klumba12, line 29: Good point need to think about interface too

klumba12 (approved):

PR #224, opened 2017-04-17
Virtualization

klumba12:


src/ng/components/cell/row-number/body.cell.row.number.html
```diff
@@ -1 +1 @@
-{{$view.pagination.current * $view.pagination.size + $cell.rowIndex + 1}}
\ No newline at end of file
+{{$cell.rowIndex + 1}}
```
titovmx, line 1: Does it change row indexes for default mode? Will it be started from 1 for each page?

etokareva (approved):

src/ng/components/cell/row-number/body.cell.row.number.html
klumba12: yep

PR #229, opened 2017-04-19
Uniq edit popup

vkorolev:
Add key unique key to popup to fix edit form overwrite bug.


src/ng/plugins/edit-form/edit.form.html
```diff
@@ -1,4 +1,4 @@
-<q-grid:popup id="q-grid-edit-form-popup">
+<q-grid:popup id="q-grid-edit-form-popup-{{$editForm.key}}">
```
klumba12, line 1: why do we need that? only one form can be opened per grid on the screen, we can use grid id instead?

PR #234, opened 2017-04-25
add number format

titovmx:
takes fraction size and applies locale pattern to number
add id column for any identificators

### DIFF
--- a/src/assets/cell.scss
+++ b/src/assets/cell.scss
@@ -114,8 +114,7 @@ thead, tfoot, tbody {
 
 tbody {
 	td {
-		&.q-grid-number,
-		&.q-grid-id {
+		&.q-grid-number {
 			text-align: right;
 		}
 
--- a/src/assets/cell.scss
+++ b/src/assets/cell.scss
@@ -114,7 +114,8 @@ thead, tfoot, tbody {
 
 tbody {
 	td {
-		&.q-grid-number {
+		&.q-grid-number,
+		&.q-grid-id {
 			text-align: right;
 		}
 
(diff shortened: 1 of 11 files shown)


src/ng/components/cell/number/body.cell.number.html
```diff
@@ -0,0 +1,4 @@
+<div ng-switch="$cell.column.format !== null">
```
klumba12: Can we use default format and get rid of ng switch?

titovmx: In any case default format changes each number with locale pattern
e.g. for en-US locale it will show zip code ```195213``` as ```195,213```
or we need in custom filter for numbers

klumba12: Lets keep it as 2 zip code is not real number i guess like an id

klumba12: Btw maybe we need to add column with type id?

klumba12: what happens when format is null?

src/assets/cell.scss
```diff
@@ -114,7 +114,8 @@ thead, tfoot, tbody {
 
 tbody {
 	td {
-		&.q-grid-number {
+		&.q-grid-number,
+		&.q-grid-id {
```
klumba12: Id should not align right

src/core/column-type/id.column.js
```diff
@@ -0,0 +1,35 @@
...
+		model: 'edit',
+		resource: column.key
+	};
+});
+
+class IdColumnModel extends DataColumnModel{
+	constructor(){
+		super('id');
+
+		this.format = '';
```
klumba12: Dont need format for id

klumba12 (approved):

PR #241, opened 2017-04-29
Editor fixes

klumba12:


etokareva (approved):

src/ng/services/grid.js
```diff
@@ -37,6 +38,14 @@ export default class Grid {
 	get Command() {
 		return Command;
 	}
+
+	get valueFactory() {
+		return valueFactory;
+	}
+
+	get labelFactory() {
+		return labelFactory();
```
titovmx: Probably there's should be ```return labelFactory;```

vkorolev (approved):

PR #246, opened 2017-05-05
Dynamic shortcut

klumba12:


src/core/edit/edit.model.js
```diff
@@ -15,7 +15,8 @@ export default class EditModel {
 			'text': 'enter',
 			'password': 'ctrl+s',
 			'number': 'ctrl+s',
-			'date': 'ctrl+s'
+			'date': 'ctrl+s',
+			'array': 'ctrl+s'
```
titovmx, line 19: Maybe is it need to add all other types?

titovmx (approved):

src/core/edit/edit.model.js
klumba12: there is $default type for other)

PR #251, opened 2017-05-11
Pass selection by keys

titovmx:
- possibility to initialize selection by keys
- selection state refactoring
- tests for multiple mode


src/core/selection/state/selection.state.js
```diff
@@ -118,6 +122,79 @@ export default class SelectionState {
 		return item;
 	}
 
+	selectByKeys(keys, unit) {
+		if (keys.length === 0) {
+			return;
+		}
+
+		if (unit == undefined) {
```
vkorolev: Use `isUndefined` instead?

src/core/selection/selection.view.js
```diff
@@ -25,10 +25,17 @@ export default class SelectionView extends View {
...
 		this.reset = commands.get('reset');
 
+		model.dataChanged.watch(() => {
+			this.selectionState = stateFactory(model);
+
+			model.selection({entries: this.selectionState.entries()});
+		});
+
 		model.sortChanged.watch(() => {
+			const entries = this.selectionState.entries();
```
klumba12: If sort is handled by server side sortChanged and dataChanged can be triggered both, consider to use viewChanged?

klumba12: And why we need to update selection on sortChanged?

src/core/selection/state/selection.state.js
```diff
@@ -118,6 +122,79 @@ export default class SelectionState {
 		return item;
 	}
 
+	selectByKeys(keys, unit) {
```
klumba12: why we need to pass here unit?

src/core/selection/state/single.selection.state.js
```diff
@@ -5,6 +5,9 @@ export default class SingleSelectionState extends SelectionState {
 		super(model);
 
 		this.item = null;
+
+		const keys = model.selection().items;
+		super.selectByKeys(keys);
```
klumba12: I'm not sure that this is good idea to have it here

src/core/selection/state/selection.state.js
klumba12: Consider to refactor this to 
```javascript
const entries = state.lookup(items);
state.select(entries)
```

src/core/selection/state/selection.state.js
```diff
@@ -118,6 +122,79 @@ export default class SelectionState {
...
+					data.rows.forEach(row => {
+						cells.push({
+							column: column,
+							row: row
+						});
+					});
+				});
+				cells.forEach(cell => {
+					const cellKey = this.key(cell);
+					const found = keys.findIndex(key => `${key.column}[${key.row}]` == cellKey) > -1;
```
klumba12: `${key.column}[${key.row}]` looks strange

titovmx: Why? It's stringified key for cell, it'll be smth like ```gender[123]``` and should be unique

src/core/selection/state/single.selection.state.js
titovmx: yes. it is. It can be placed to parent contsructor or be used outside of state. I'll think about that

src/core/selection/state/selection.state.js
titovmx: unit is passed for mix mode case, when unit depends on each item

src/core/selection/state/selection.state.js
klumba12: because if it was smth like `key => this.key(key) === cellKey` it wasn't strange :)
btw please use ===

klumba12 (approved):

PR #253, opened 2017-05-16
Get rid of applyFactory in CORE

klumba12:


src/view/services/command.manager.js
```diff
@@ -0,0 +1,12 @@
+export class CommandManager {
+	constructor(apply) {
+		this.apply = apply;
+	}
+
+	execute(commands) {
+		// First we need to get list of executable commands, cause execution of prev command can
```
vkorolev, line 7: I think it's correct behavior, when a command affects possibility to execute of the next command. Maybe we need to execute them sequentially?

src/view/services/command.manager.js
```diff
@@ -0,0 +1,12 @@
+export class CommandManager {
```
vkorolev, line 1: Personally I don't like this name:)
For me it seems to be something like execution environment. What about Executor or maybe a Sandbox?

src/view/services/command.manager.js
klumba12: What do you mean by sequentially?
Real example for enter shortcut:
```click enter: open popup -> set popup open = true```
```click enter: if popup open === ture -> close it```
So popup for the user won't open. (it will open and close by enter hit)
We just need to manipulate with command sets to avoid that, but this is kind of big refactoring

src/view/services/command.manager.js
klumba12: No it's really manage command calls, so that why it has name like that.

src/view/services/command.manager.js
vkorolev: As I understood, `commandManager.execute` is triggered only from `onKeyDown` method of Shortcut instance. Why `enter` is triggered twice per single key press?
I think this approach may introduce new bugs in future.

src/view/services/command.manager.js
vkorolev: It doesn't manage commands in any way: it doesn't add or delete them. It only filters and executes them within some external context (apply)

vkorolev (approved):

PR #110, opened 2017-02-10
Bower integration

vkorolev:

PR #267, opened 2017-06-01
Export

etokareva:


src/core/export/export.common.js
```diff
@@ -0,0 +1,30 @@
+import {isObject, isArray} from '@grid/core/services/utility';
+
+const flattenObject = (obj, separator = ' ,') => {
+	const result = {};
+	for (let prop in obj) {
```
titovmx: You can use 
```
const props = Object.keys(obj);
props.forEach(prop => {
    // first level
    ...
}
```
instead of
```
for (let prop in obj) {
    // first level
    if (obj.hasOwnProperty(prop)) {
       // second level
        ...
    }
}
```

It'll have the same behavior and can help to reduce deep nesting

src/core/export/export.common.js
```diff
@@ -0,0 +1,30 @@
...
+		if (obj.hasOwnProperty(prop)) {
+			if (isObject(obj[prop]) && !isArray(obj[prop])) {
```
titovmx:
```
if (isArray) {
    ...
} else if (isObject) {
    ...
} else {
    ...
}
```

imo just a bit easier :)

titovmx:
I think ```/dist``` folder shouldn't be in the diff

etokareva:
I'm agree. I've commented about this earlier: https://github.com/qgrid/ng/pull/110
But nothing has been changed.
This is the only reason, why I haven't included dist folder in gitignore.

titovmx:
@etokareva this folder is neccessary for release branch
Probably we should keep different .gitignore for them

etokareva:
@titov-max let see what @klumba12  will say

src/core/export/export.common.js
etokareva: Thnx :-) Get the idea. But if you don't mind, I'll use Object.entries

dist/qgrid.css
```diff
@@ -0,0 +1,1547 @@
+.q-grid-popup {
```
klumba12: dist folder shouldn't be in master branch

src/core/export/csv.js
```diff
@@ -2,12 +2,12 @@ import {getFactory as valueFactory} from '../services/value';
 
 const DELIMETER = ',';
 
-function escape(value) {
+const escape = (value) => {
```
klumba12: why? =) i like function notation more in this case

src/core/export/export.common.js
```diff
@@ -0,0 +1,27 @@
+import {isObject, isArray} from '@grid/core/services/utility';
+
+const flattenObject = (obj, separator = ' ,') => {
```
klumba12: what it does?  can't understand

src/core/export/csv.js
etokareva: ok, I'll change then))) I'm too clever by half :-)

src/core/export/export.common.js
etokareva: It turns tree structure into flat

package.json
```diff
@@ -48,11 +48,14 @@
 		"eslint-plugin-angular": "^1.6.1",
 		"extract-text-webpack-plugin": "^2.0.0-beta.5",
 		"file-loader": "^0.9.0",
+		"file-saver": "^1.3.3",
```
vkorolev: FileSaver, jspdf and xlsx should be put to dependencies

etokareva: @klumba12 said yesterday to put them in devDependencies

.gitignore
```diff
@@ -9,5 +9,5 @@
 node_modules
 
 #bundles
-dist/vendor.*
+dist
```
titovmx: I'd like to draw attention that this change will break work of release script

It's neccessary order when you merge the pr:

1) add dist to .gitignore in master (merge pr)
2) manually merge master to release
3) remove dist from .gitignore in release
4) manually push release

titovmx: Hmm and then all users must manually pull release before using of script

etokareva: too many complexities. Let me just undo this change

klumba12 (approved):

PR #277, opened 2017-06-06
Fix infinite scroll: add fetch model

klumba12:


demo/pages/infinite-scrolling/index.js
```diff
@@ -16,14 +16,12 @@ export default function Controller($http, qgrid) {
...
-			const size = paginationState.size;
-			const current = paginationState.current;
-			const start = current * size;
+			const skip = model.fetch().skip;
+			const take = model.pagination().size;
 
 			$http.get('data/people/100.json')
 				.then(response =>
-					next(rows = rows.concat(response.data.slice(start, start + size))));
+					next(rows = rows.concat(response.data.slice(skip, skip + take))));
```
vkorolev, line 24: Maybe we can move it to the new pipe? something like accumulator which would work only with virtual mode

vkorolev (approved):

PR #285, opened 2017-06-13
Selection refactoring

klumba12:


src/view/components/grid/grid.js
```diff
@@ -66,16 +69,29 @@ export class Grid extends RootComponent {
 		});
 	}
 
-	applyFactory(gf) {
+	applyFactory(gf = null, mode = 'async') {
```
titovmx, line 72: what is gf, lf?

klumba12: it;s like of globalF and localF, however i'm still thinkinkg how to get rid of that applyFactory

titovmx (approved):

PR #288, opened 2017-06-14
Selection & popup bug fixes

klumba12:


src/view/services/grid.js
```diff
@@ -2,7 +2,7 @@ import {Model, Log, Command, Action} from '@grid/core/infrastructure';
 import {GridService} from '@grid/core/services';
 import {Pipe} from '@grid/core/pipe';
 import {PipeUnit} from '@grid/core/pipe/units';
-import {noop} from '@grid/core/utility';
+import {noop, identity} from '@grid/core/utility';
```
vkorolev, line 5: Approve, but why do we need to export these utility functions?

vkorolev (approved):

src/view/services/grid.js
klumba12: Because they have some default model values
If user want to reset model to default he'll be forced to use that

PR #292, opened 2017-06-15
Virtualization fixes

klumba12:
Reset scroll on view changed in virtual mode;

### DIFF
--- a/src/core/dom/virtual/cell.js
+++ b/src/core/dom/virtual/cell.js
@@ -1,10 +1,41 @@
 import {Cell} from '../cell';
+import * as columnService from '../../column/column.service';
+import {Cell as CellModel} from '../../cell';
 
 export class VirtualCell extends Cell {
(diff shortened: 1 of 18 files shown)


src/core/dom/virtual/cell.js
```diff
@@ -1,10 +1,41 @@
...
+					columnIndex: columnIndex,
+					row: rows[rowIndex],
+					column: columns[columnIndex]
+				};
+
+				return new CellModel(model);
+			}
+		}
+
+		return null;
```
vkorolev, line 38: Maybe return null-object here?

titovmx (approved):

etokareva (approved):

src/core/dom/virtual/cell.js
klumba12: I was thinking about that and decided to return null, cause model is a lot by validation

PR #295, opened 2017-06-20
Import

etokareva:


demo/pages/import/index.html
```diff
@@ -0,0 +1,19 @@
+<q-grid model="$ctrl.model"
+        edit-mode="cell"
+        selection-mode="range">
+	<q-grid:columns ng-if="$ctrl.model.data().rows.length">
```
titovmx: Why do we need in this ng-if expression?
Besides, I don't see that ```model``` put in ```$ctrl```

src/core/export/csv.js
```diff
@@ -15,8 +15,10 @@ export class Csv {
 		const values = [];
 		let head = [];
 		for (let column of columns) {
-			values.push(valueFactory(column));
-			head.push(escape(column.title));
+			if (column.key !== 'rowNumber') {
```
titovmx: Maybe check type will be better. 
And what's about row-indicator column? Should it be here too?

src/core/import/csv.js
```diff
@@ -0,0 +1,53 @@
+function arrayToObj(array) {
```
klumba12: please add tests for that file

src/core/import/xml.js
```diff
@@ -0,0 +1,70 @@
+const types = {
```
klumba12: add tests

src/core/navigation/navigation.view.js
```diff
@@ -15,15 +15,23 @@ export class NavigationView extends View {
 		this.shortcutOff = shortcut.register('navigation', navigation.commands);
 
 		this.blur = new Command({
-			execute: (row, column) => table.body.cell(row, column).removeClass(`${GRID_PREFIX}-focus`),
+			execute: (row, column) =>{
+				table.body.cell(row, column).removeClass(`${GRID_PREFIX}-focus`);
```
klumba12: this should be done outside the grid in excel example, if it's not possible we will add this future,
for now try to use style API

src/core/services/title.js
```diff
@@ -0,0 +1,23 @@
+function firstRowTitle(index, row) {
+	return row[index];
+}
+function numericTitle(index) {
+	return index;
+}
+function alphabeticalTitle(index) {
```
klumba12: alphaTitle is ok

src/plugins/import/import.html
```diff
@@ -0,0 +1,3 @@
+<div class="q-grid-export">
```
klumba12: q-grid-export?

src/plugins/import/import.html
```diff
@@ -0,0 +1,3 @@
...
+	<button type="button" ng-if="$import.upload.canExecute()">Upload file</button>
```
titovmx: Probably ```ng-click``` should be here

demo/pages/import/index.html
etokareva: thanks, forgot to remove

src/core/export/csv.js
etokareva: Maybe.
No, it shouldn't

titovmx: Another one service column is ```pad``` column. Will it be exported?

src/plugins/import/import.html
etokareva: yes, this is mistake)

src/core/export/csv.js
etokareva: Theme with columns for export should be discussed more, I think

src/core/column-list/column.list.generate.js
```diff
@@ -1,15 +1,16 @@
 import {startCase} from '../utility';
 import {compile, getType} from '../services';
 
-export function generate(rows, columnFactory, deep = true) {
-	if (!rows || rows.length === 0) {
+export function generate(context) {
+	const deep = context.deep || true;
+	if (!context.rows || context.rows.length === 0 || !context.columnFactory) {
```
klumba12: please use default object for context

demo/pages/import/index.md
```diff
@@ -0,0 +1,48 @@
...
+<thead>
+	<tr>
+		<th>Parameter</th>
+		<th>Type</th>
+		<th>Description</th>
+	</tr>
+</thead>
+<tbody>
+    <tr>
+        <td>header</td>
```
klumba12: head: 'alpha' | 'numeric' | 'default'

src/core/column-list/column.list.generate.js
klumba12: title by default shouldBe startCase
signature is: title(propertyPath, graph, index)

src/core/export/csv.js
klumba12: column has a class property, you can export only class === 'data' for example

src/core/import/csv.js
```diff
@@ -0,0 +1,53 @@
...
+	const result = {};
+	for (let i = 0; i < array.length; i++) {
+		result['' + i] = array[i];
+	}
+	return result;
+}
+
+function traverse(row) {
+	const result = [];
```
klumba12: need to refactor this

src/core/import/json.js
```diff
@@ -0,0 +1,11 @@
+import {isArray} from '@grid/core/utility';
+
+export class Json {
+	read(data) {
+		const rows = JSON.parse(data);
+		if (isArray(rows)) {
+			return rows;
+		}
+		return [];
```
klumba12: [rows]

src/core/import/xml.js
```diff
@@ -0,0 +1,70 @@
...
+	return children.length > 1;
+}
+function isElement(xml) {
+	return xml.nodeType === types.ELEMENT_NODE;
+}
+function isTextContainer(xml) {
+	return xml.nodeType === types.ELEMENT_NODE && !xml.children.length && xml.childNodes.length;
+}
+
+function generateArray(child) {
```
klumba12: not readable name, cant understand what it does

src/core/import/xml.js
```diff
@@ -0,0 +1,70 @@
...
+			result.push(obj);
+		}
+		child.setAttribute('visited', '');
+	}
+	return result;
+}
+function isVisited(xml) {
+	return xml.getAttribute('visited');
+}
+function parse(xml, obj, tagName) {
```
klumba12: the same parse doesn't return anything so don't understand what it does

src/plugins/import/import.js
```diff
@@ -0,0 +1,64 @@
...
+				const input = doc.createElement('input');
+				input.type = 'file';
+				input.className = 'ng-hide';
+				element.appendChild(input);
+				input.click();
+			}
+		});
+	}
+
+	handleFile(e) {
```
klumba12: rename it

src/plugins/export/pdf.js
```diff
@@ -1,6 +1,6 @@
 import jsPDF from 'jspdf';
```
klumba12: don't import this files directly, use global namespace for them, also add them to vender list in webpack config.
now they are written to dist, also in documention please add that user should import this vendor libraries

src/core/export/csv.js
titovmx: ```class``` property is really good. But maybe it should be defined as 'data' in DataColumnModel? I'm confused a bit that it's declared in base class

src/core/import/xml.js
etokareva: it creates array

src/core/import/xml.js
etokareva: it changes the result object, which we pass from the outside using xml object

src/plugins/export/pdf.js
etokareva: it's already in webpack config. It has been added in my previous pull request

src/plugins/import/import.js
etokareva: 'load ' will be okay?

src/core/import/json.js
etokareva: this won't work in case of return [rows] if rows is array

src/core/column-list/column.list.generate.js
etokareva: which object is default?

src/core/export/csv.js
etokareva: Actually, I found out that for column type: "row-number" class is equal to "data"

demo/pages/excel/index.js
```diff
@@ -0,0 +1,49 @@
...
+				rows.push(row);
+			}
+
+			next(rows);
+		}]
+			.concat(qgrid.pipeUnit.view)
+	});
+
+	model.style({
+		// row: (row, context) => {
```
klumba12: please, remove these comments

src/core/column-list/column.list.generate.js
```diff
@@ -1,15 +1,21 @@
...
-	if (!rows || rows.length === 0) {
-		return [];
+export function generate(settings) {
+	const context = assignWith({
+		deep: true,
+		rows: [],
+		columnFactory: () => new TextColumnModel(),
+		title: startCase
+	}, settings);
+	if (context.rows && context.rows.length) {
```
klumba12: don't need to check context.rows

demo/pages/excel/index.js
vkorolev: Consider to uncomment or delete this snippet

src/core/export/export.service.js
```diff
@@ -23,5 +23,5 @@ function flattenObject(obj, separator = ', ') {
 }
 
 export {
-	flattenObject
+	flatView
```
vkorolev, line 26: **export** service exposes single function called `flatView`. Consider to rename file to flatten or like this

src/core/import/csv/csv.spec.js
```diff
@@ -0,0 +1,34 @@
+import {Csv} from './csv';
+
+const csv = new Csv();
+
+describe('csv parser', () => {
+	it('should return empty array', () => {
```
vkorolev: should return empty string when input is not defined

etokareva: I think it should return array of rows

src/core/import/csv/csv.spec.js
```diff
@@ -0,0 +1,34 @@
...
+		expect(csv.read()).to.eql([]);
+	});
+	it('should return empty array when passed empty string', () => {
+		expect(csv.read('')).to.eql([]);
+	});
+
+	it('should return array of rows', () => {
+		const text = `Name First,Name Last,Gender,Birthday,Contact Address Street,Contact Address Zip,Contact Address City,Contact Address State,Contact Email,Contact Region,Contact Phone,Likes,Member Since
```
vkorolev: imo this test data is too verbose. i think 2x2 table would be enough

etokareva: ok

src/core/export/json.js
```diff
@@ -14,34 +14,17 @@ function rewriteHeaders(rows, titles) {
...
-
-			return rewriteHeaders(result, titles);
+		for (let row of rows) {
+			result.push(flatView(row));
+		}
+		for (let column of columns) {
+			titles.push(column.title);
 		}
 
+		return rewriteHeaders(result, titles);
```
klumba12: rename it to updateTitles

src/core/import/csv/csv.spec.js
vkorolev: sorry, of course.
i meant that you should have more detailed sentence 
`'should return empty array wnhen input is not defined'`

src/core/import/csv/csv.spec.js
vkorolev: It's better to have multiple small tests which would check parser's edge cases

src/core/export/xml.js
```diff
@@ -53,22 +34,14 @@ function objToXml(obj) {
...
-	headers.push('</head>');
-	return headers.join('');
-}
 
 export class Xml {
-	write(rows, columns) {
-		const result = [`${begin}${schema}`, ...formHead(columns), '<body>'];
+	write(rows) {
+		const result = [begin];
 		for (let row of rows) {
```
klumba12, line 41: rows.map(objToXml) looks pretty doesn't it? :)

src/plugins/import/import.js
```diff
@@ -0,0 +1,64 @@
...
+const Plugin = PluginComponent('import');
+class Import extends Plugin {
+	constructor() {
+		super(...arguments);
+		const element = this.$element[0];
+		const doc = element.ownerDocument;
+		this.eventListener = new EventListener(element, new EventManager(this));
+		this.upload = new Command({
+			execute: () => {
+				const input = doc.createElement('input');
```
klumba12: can we have in core upload.js with that code?

src/core/import/csv/csv.js
```diff
@@ -0,0 +1,128 @@
...
+
+				continue;
+			}
+
+			if (c === CharReader.eof) {
+				reader.read();
+
+				if (line.length > 0 || term.length > 0) {
+					line.push(term);
+					if (line.length > 0) {
```
titovmx: I think this condition is redundant

src/core/import/xml/xml.spec.js
```diff
@@ -0,0 +1,126 @@
...
+		expect(xml.read()).to.eql([]);
+	});
+
+	it('should return empty array when passed empty string', () => {
+		expect(xml.read('')).to.eql([]);
+	});
+
+	it('should return array of rows', () => {
+		const text = `<?xml version="1.0" encoding="UTF-8"?>
+		<root>
```
titovmx, line 30: can it be placed to separate file or reduced?

src/view/components/head/head.html
```diff
@@ -1,18 +1,19 @@
...
+	    q-grid:on-drop="$view.head.drop.execute($event)"
+	    q-grid:can-drop="$view.head.drop.canExecute($event)"
+	    q-grid:can-drag="$view.head.drag.canExecute($event)"
+	    q-grid:resizable="$view.head.resize.canExecute($event)"
+	    q-grid:resize="$column.model.key"
+	    q-grid:resize-path="columns"
+	    q-grid:can-resize="$view.head.resize.canExecute($event)"
+	    q-grid-core:th>
 	</th>
+	<th ng-if="$view.head.rows.length === 1 && $view.body.rows.length === 0"></th>
```
titovmx: why do we need in this ```th```?

src/core/export/export.service.js
klumba12: Its ok name i think no need to make separate file in this case

src/view/components/head/head.html
klumba12: +1

etokareva: Actually, now we don't need this :-)

src/core/export/xml.js
etokareva: I don't know how to wrap object in row tag in that case

src/core/import/csv/csv.js
etokareva: I think you are right

src/core/import/xml/xml.spec.js
etokareva: reduced

klumba12 (approved):

PR #296, opened 2017-06-21
Selection command manager

titovmx:


src/core/selection/selection.command.manager.js
```diff
@@ -0,0 +1,19 @@
...
+export class SelectionCommandManager extends CommandManager {
+	constructor(model, manager) {
+		super();
+
+		this.model = model;
+		this.manager = manager;
+	}
+
+	keyDown(f) {
+		this.manager.keyDown(f);
```
klumba12: Return?

klumba12: Maybe add composite command manager as base class to avoid that kind of errors?

klumba12 (approved):

PR #301, opened 2017-06-26
Filter row

titovmx:


src/core/column-type/filter.row.column.js
```diff
@@ -0,0 +1,32 @@
...
+
+export class FilterRowColumnModel extends ColumnModel {
+	constructor(model) {
+		super();
+
+		Object.assign(this, model);
+
+		this.key = `$filter.row.${model.key}`;
+		this.type= 'filter-row';
+		this.class = 'filter';
```
klumba12: setup it as 'control' not 'filter' for now

src/core/head/head.view.js
```diff
@@ -76,5 +108,11 @@ export class HeadView extends View {
 		Log.info('view.head', 'invalidate');
 
 		this.rows = model.view().columns;
+
+		if (model.filter().unit === 'row') {
```
klumba12: what happen when user setup filter.unit to default? does it dissapering now?

src/core/column-type/column.model.js
```diff
@@ -10,7 +10,7 @@ export class ColumnModel {
 		this.pin = null;
 		this.origin = 'specific';
 		this.source = 'user';
-		this.class = 'data';
+		this.class = '';
```
klumba12: Doesnt it break user defined columns?

titovmx: I didn't see direct problems.
User can define both data and control columns, so in my opinion he has to define class himself when he need.

klumba12: I think when user create own column he shouldn't think about it class, class is done for internal uses to the most of time

src/core/column/column.view.js
```diff
@@ -46,6 +46,13 @@ export class ColumnView extends View {
 				service.invalidate('column.view', e.changes, PipeUnit.column);
 			}
 		});
+
+		model.filterChanged.watch(e => {
```
klumba12: i think this should be in headview itself without calling pipe in other case u use two different approaches: pipe and headview.invalidate

klumba12 (approved):

PR #302, opened 2017-06-26
Layer layout navigation node pagination

munkk:


src/core/body/body.view.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export declare interface IGetValueFactory{
+	(column: ColumnModel): IGetResult;
+}
+
+export declare class BodyView extends View {
+	constructor(public model: Model, public table: Table);
+
+	table: Table;
+	rows: any[];
```
vkorolev, line 16: not any, maybe RowModel?

src/core/column-list/column.list.model.d.ts
```diff
@@ -0,0 +1,16 @@
+import {ColumnModel} from "../column-type/column.model";
+
+export declare enum Generation{
```
vkorolev, line 3: Better make it a type like this:
`type Generation = 'deep' | 'shallow';`
https://basarat.gitbooks.io/typescript/docs/types/literal-types.html

vkorolev:
I think I saw some these changes in previous PR's. Can you please merge these branches or change PR destination to simplify review? 
Tell us if need any help with this :)

src/core/column-list/column.list.model.d.ts
munkk: thx)

PR #326, opened 2017-06-30
Master details

etokareva:


demo/pages/master-details/index.html
```diff
@@ -0,0 +1,16 @@
+<q-grid rows="$ctrl.masterRows"
```
klumba12: can we use <div flex=50> as a grid containers and get rid of .multigrid?

demo/pages/master-details/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+	this.selectionChanged = (e) => {
+		this.selection = e.state.items;
+		if (this.selection.length) {
+			const selection = this.selection[0].likes;
+			this.likes = selection;
+			$http.get('data/people/10.json')
+				.then(response => {
+					const detailsRows = response.data.filter(r => {
+						return selection.every(val => r.likes.indexOf(val) >= 0);
```
klumba12: rename selection to likes
rename val to like

klumba12 (approved):

PR #336, opened 2017-07-05
Edit improvements

etokareva:


src/core/edit/edit.cell.view.js
```diff
@@ -49,9 +52,17 @@ export class EditCellView {
 					}
 					else {
 						cell = model.navigation().cell;
+						const code = this.shortcut.keyCode;
+						if (code && code.length > 0) {
+							const reserved = Array.from(this.shortcut.codeMap.values()).some(val => val === code);
```
klumba12: add to shortcut isControl method instead of that

src/core/edit/edit.cell.view.js
```diff
@@ -49,9 +52,17 @@ export class EditCellView {
...
+							if (!reserved) {
+								this.memory = code;
```
klumba12: why we need memory?

src/core/edit/edit.cell.view.js
```diff
@@ -49,9 +52,17 @@ export class EditCellView {
 					}
 					else {
 						cell = model.navigation().cell;
+						const code = this.shortcut.keyCode;
+						if (code && code.length > 0) {
```
klumba12: in what cases length is 0?

src/core/edit/edit.cell.view.js
```diff
@@ -49,9 +52,17 @@ export class EditCellView {
...
+							const reserved = Array.from(this.shortcut.codeMap.values()).some(val => val === code);
+							if (!reserved) {
+								this.memory = code;
+								cell.value += code;
+							}
+						}
 					}
 
-					if (cell && model.edit().enter.execute(this.contextFactory(cell, cell.value, cell.label)) !== false) {
+					if (model.edit().enter.execute(this.contextFactory(cell, cell.value, cell.label)) !== false) {
```
klumba12: why cell is excluded from if? it can be null

src/core/edit/edit.cell.view.js
```diff
@@ -201,19 +221,30 @@ export class EditCellView {
...
+	}
+
 	commitShortcut() {
 		const model = this.model;
-		const commitShortcuts = model.edit().commitShortcuts;
+		const shortcuts = model.edit().commitShortcuts;
 		const cell = this.editor.cell || model.navigation().cell;
-		if (cell && commitShortcuts.hasOwnProperty(cell.column.type)) {
-			return commitShortcuts[cell.column.type];
+		if (cell && shortcuts.hasOwnProperty(cell.column.type)) {
```
klumba12: please add editor check here

src/core/edit/edit.cell.view.js
```diff
@@ -201,19 +221,30 @@ export class EditCellView {
...
+		return shortcuts['$default'];
 	}
 
-	get options() {
-		return this.editor.options;
+	enterShortcut() {
+		const model = this.model;
+		const shortcuts = model.edit().enterShortcuts;
+		const cell = this.editor.cell || model.navigation().cell;
+		if (cell && shortcuts.hasOwnProperty(cell.column.type)) {
```
klumba12: please add editor check here

src/core/edit/edit.cell.view.js
etokareva: I need a flag to know about starting the input modifications with any key (to add it to the cell value, and to remove it on cancel)

src/core/behavior/shortcut.js
```diff
@@ -0,0 +1,63 @@
...
+export class Shortcut {
+	constructor(manager) {
+		this.manager = manager;
+	}
+	get keyCode() {
+		return Shortcut.keyCode;
+	}
+
+	get isControl() {
+		return Array.from(codeMap.values()).some(val => val === Shortcut.keyCode);
```
klumba12: can u add here another control values like ctrl F1 etc.?
and please add argument isControl(code),
Thanks!

src/core/behavior/shortcut.manager.js
```diff
@@ -0,0 +1,108 @@
+import {flatten, isFunction} from '../utility';
+import {Command} from './command';
+
+export class ShortcutManager {
+	constructor() {
+		this.managerMap = new Map();
+		this.keyCode = '';
```
klumba12: Why we need to store here keyCode, looks wrong place here

src/core/edit/edit.cell.view.js
```diff
@@ -49,6 +52,11 @@ export class EditCellView {
 					}
 					else {
 						cell = model.navigation().cell;
+						const code = this.shortcut.keyCode;
+						if (cell && !this.shortcut.isControl) {
+							this.memory = code;
```
klumba12: memory? for what we need to have it?

src/core/edit/edit.cell.view.js
```diff
@@ -49,6 +52,11 @@ export class EditCellView {
...
+							cell.value += code;
```
klumba12: i think it should assign code not append like in excel

src/core/edit/edit.cell.view.js
```diff
@@ -109,7 +117,16 @@ export class EditCellView {
 					}
 
 					cell = cell || this.editor.cell || model.navigation().cell;
-					if (cell && model.edit().cancel.execute(this.contextFactory(cell, this.value, this.label)) !== false) {
+					let label = this.label;
```
klumba12, line 121: don't understand this code, why we need to anything on cancel? it should just close editor

src/core/edit/edit.cell.view.js
klumba12: why u need to know it ? it's not allowed to cell.value until commit is triggred, for that this.value was introduced

etokareva: Because I have to add and remove letter, which opens edit-mode

klumba12 (approved):

PR #338, opened 2017-07-07
fixed UI directions for sorting (up for asc and down for desc)

peaceful-djony:


src/themes/material/templates/head.cell.text.html
```diff
@@ -5,9 +5,9 @@
 		  ng-class="{'q-grid-active': $view.sort.order($cell.column) >= 0}"
 		  ng-if="$cell.column.canSort"
 		  ng-switch="$view.sort.direction($cell.column)">
-		<md-icon class="q-grid-asc" ng-switch-when="asc">arrow_downward</md-icon>
-		<md-icon class="q-grid-desc" ng-switch-when="desc">arrow_upward</md-icon>
-		<md-icon ng-switch-default ng-if="$view.sort.hover">arrow_downward</md-icon>
+		<md-icon class="q-grid-asc" ng-switch-when="asc">arrow_upward</md-icon>
```
etokareva, line 8: Why we need to change the icons?

etokareva: When data sorted is asc our hint arrow should be 'arrow_downward' to show user that he can click and sort data in desc order

klumba12 (approved):

PR #346, opened 2017-07-12
Add area, toggle and select for selection model

vkorolev:


etokareva (approved):

src/core/selection/selection.model.js
```diff
@@ -1,12 +1,16 @@
 import {Resource} from '../resource';
 import {identity} from '../utility';
+import {Command} from '../command';
 
 export class SelectionModel {
 	constructor() {
 		this.resource = new Resource();
 		this.unit = 'cell';//row|cell|column|mix
 		this.mode = 'single';//single|multiple|range
 		this.items = [];
```
titovmx, line 10: I think, type definitions should be updated too

vkorolev: good point, thanks

klumba12 (commented):
Please add bindings to the grid component, and maybe get rid of select command and use toggle everywhere? With kind= toggle or select

PR #356, opened 2017-07-25
Correct navigation with tabs through form controls

klumba12:


src/core/dom/view.js
```diff
@@ -24,7 +24,7 @@ export class View extends Unit {
 	}
 
 	focus() {
-		const elements = this.getElementsCore('table');
+		const elements = this.getElementsCore('body');
```
titovmx, line 27: Shouldn't it be ```this.getElementCore()```?

vkorolev (approved):

src/core/dom/view.js
klumba12: no =)

PR #358, opened 2017-07-27
moved core d.ts from Ng2 to Ng1

munkk:


src/core/body/body.model.d.ts
```diff
@@ -2,7 +2,7 @@ import {Resource} from '../resource/resource';
 import {Cache} from '../infrastructure/cache';
 
 export declare class BodyModel {
-	constructor();
-	resource: Resource;
-	cache: Cache;
+  constructor();
```
vkorolev, line 5: You have some files formatted with tabs and some - with spaces. Can you format them in a single way?

klumba12 (approved):

PR #386, opened 2017-08-11
Event test

munkk:


src/core/infrastructure/event.spec.js
```diff
@@ -1,4 +1,39 @@
...
+	});
+
+	let foo = function (value, func) {
+		func();
+		++window.testValue;
+	};
+
+	describe('on/emit', () => {
+
+		it('should return 1 in both cases', () => {
```
vkorolev: Consider to use [Chai spies](http://chaijs.com/plugins/chai-spies/) .
In this case you won't need to test side effects in window object. It'll be look like this:
```javascript
const foo = chai.spy();
event.on(foo);
expect(foo).to.have.been.called();
```

PR #391, opened 2017-08-15
Command manager test

munkk:


src/core/command/command.manager.spec.js
```diff
@@ -0,0 +1,40 @@
+import {CommandManager} from './command.manager';
+
+describe('CommandManager', () => {
+	let arrayOfCommand = [],
```
klumba12: rename it to commands

klumba12 (approved):

PR #392, opened 2017-08-16
Composite manager test

munkk:


src/core/command/composite.command.manager.spec.js
```diff
@@ -0,0 +1,42 @@
+import {CompositeCommandManager} from './composite.command.manager';
+import {CommandManager} from './command.manager'
+
+describe('CompositeCommandManager', () => {
+	let arrayOfCommands = [],
```
klumba12: rename it to commands

klumba12 (approved):

PR #395, opened 2017-08-16
Added data column generation to the data pipe, pagination as last pipe, Fix dts files 

klumba12:


src/core/infrastructure/event.d.ts
```diff
@@ -1,15 +1,15 @@
...
-  (): void;
+	(): void;
 }
 
 export declare class Event {
-  constructor(e?: () => any);
+	constructor(e?: () => any);
 
-  on(f: IFunc): IReturn;
+	on(handler: IFunc): IReturn;
```
titovmx, line 10: maybe rename ```IFunc``` to something like ```IEventHandler```?

src/core/infrastructure/event.d.ts
```diff
@@ -1,15 +1,15 @@
...
 
-  watch(f: IFunc): void;
+	watch(handler: (e: any, off?: () => void) => void): () => void;
```
titovmx, line 12: really hard to read )

titovmx: what's difference between this and ```IFunc```?

klumba12: this is Func of Func)

klumba12: more complicated then just IFunc

klumba12: will make better for ng2 later

titovmx (approved):

PR #400, opened 2017-08-16
Unit test

munkk:


src/core/dom/unit.spec.js
```diff
@@ -0,0 +1,37 @@
...
+		left: 0,
+		right: 0,
+		top: 0,
+		bottom: 0,
+		width: 0,
+		height: 0
+	};
+
+	describe('getElement', () => {
+		let elem = unit.getElement();
```
klumba12, line 16: This is private

src/core/dom/unit.spec.js
```diff
@@ -0,0 +1,37 @@
+import {Unit} from './unit';
+import {FakeElement} from './fake';
+
+describe('Unit', () => {
+	let unit = new Unit();
```
klumba12, line 5: Please, also write test where u pass element to unit ctor.

PR #410, opened 2017-08-21
Cast factory test + cast factory.js fix

munkk:


src/core/expression/cast.factory.js
```diff
@@ -1,4 +1,5 @@
 import {AppError} from '../infrastructure';
+import * as Convert from '../services/convert';
 
 export function castFactory(r) {
 	const rt = typeof r,
```
klumba12: there is sence to use Convert.getType here too

vkorolev: Please, add also failing test which would be fixed by `Convert.getType`

src/core/expression/cast.factory.spec.js
```diff
@@ -0,0 +1,34 @@
...
+		it('should convert "123" to number', () => {
+			let foo = CastFactory.castFactory('123');
+			let result = foo(1);
+			expect(result).to.equal(123);
+		});
+
+		it('should convert string to Date', () => {
+			let foo = CastFactory.castFactory('2015-05-05');
+			let result = foo(new Date(2017));
+			expect(result instanceof Date).to.equal(true);
```
vkorolev: Maybe `expect(result).to.be.an.instanceof(Date)` ?

src/core/expression/cast.factory.spec.js
```diff
@@ -0,0 +1,34 @@
+import * as CastFactory from './cast.factory';
+
+describe('CastFactory', () => {
+	describe('castFactory', () => {
+		it('should return 123', () => {
+			let foo = CastFactory.castFactory(123);
```
klumba12, line 6: add tests for boolean and for constructros like new Number(123)

src/core/expression/cast.factory.js
```diff
@@ -1,29 +1,36 @@
 import {AppError} from '../infrastructure';
 
+function getType(type) {
+	return {}.toString.call(type).slice(8, -1);
```
klumba12: instead of 8 please write 'object '].length or something simlar for readability

klumba12 (approved):

PR #411, opened 2017-08-22
Expression build test

munkk:


src/core/expression/expression.build.spec.js
```diff
@@ -0,0 +1,31 @@
+import * as ExpressionBuild from './expression.build';
+
+describe('Expression Build', () => {
+	let test1 = {
+		key: {
```
klumba12, line 5: rename test1 to filterBy1,
with next schema
{
 field1: { items: ['foo', ['bar']]},
filed2: {expression: 'nextExpression'}
}

klumba12 (approved):

PR #412, opened 2017-08-23
Predicate.visitor.test

munkk:


src/core/expression/predicate.visitor.spec.js
```diff
@@ -0,0 +1,384 @@
...
+				},
+				left: {
+					kind: 'condition',
+					op: 'equals',
+					right: 123,
+					left: 'value'
+				}
+			};
+			let predicateVisitor = new PredicateVisitor(valueFactory);
+			let foo = predicateVisitor.visitGroup(group);
```
vkorolev: Entry point for all visitors is `visit` method. Internally it determines, how to handle it.
So you can change it to `predicateVisitor.visit(group)` everywhere.

vkorolev: Also, there's a minor error in typing for `expression.visitor`. `visit(item: IRoot, depth: number): void;` - depth is optional here.

vkorolev (approved):

PR #413, opened 2017-08-23
Predicate visitor test

munkk:


src/core/expression/predicate.visitor.spec.js
```diff
@@ -0,0 +1,239 @@
...
+			let condition = {
+				kind: 'condition',
+				op: 'between',
+				right: [10, 20],
+				left: 'value'
+			};
+			let predicateVisitor = new PredicateVisitor(valueFactory);
+			let foo = predicateVisitor.visit(condition);
+			let res = foo(15);
+			expect(res).to.equal(true);
```
titovmx, line 107: can you make separate test for each case? I mean one theory contains one expectation
smth like
```it('check equality with op = "between" inside range')```
```it('check equality with op = "between" on bounds of range ')```
```it('check equality with op = "between" outside range')```

File: src/core/expression/predicate.visitor.spec.js [hunk 2]
```diff
@@ -0,0 +1,239 @@
...
+				kind: 'condition',
+				op: 'like',
+				right: 'SOME',
+				left: 'value'
+			};
+			let predicateVisitor = new PredicateVisitor(valueFactory);
+			let foo = predicateVisitor.visit(condition);
+			let res = foo('findSomeString');
+			expect(res).to.equal(true);
+			let res2 = foo('findSomeString');
```
titovmx: Seems that you test the same again :) and there will be better to create own test for this case also

titovmx (approved):

klumba12 (approved):

PR #414, opened 2017-08-23
Group enh

klumba12:


etokareva (approved):

demo/pages/grouping-model/index.js
```diff
@@ -1,9 +1,36 @@
...
+			context.class('group', {background: '#E3F2FD'});
+		}
+	};
+
+	this.rowOptions = {
+		trigger: 'click',
+		actions: [
+			new qgrid.Action(
+				new qgrid.Command({
+					canExecute: () => {
```
etokareva, line 20: is it already true by default?

src/core/infrastructure/range.d.ts
```diff
@@ -0,0 +1,6 @@
+export declare class Range {
+	constructor(start: Number, end: Number);
```
titovmx, line 2: Do you need Number object type exactly here?

src/core/column-type/column.model.view.js
```diff
@@ -19,8 +19,10 @@ TemplatePath.register('custom-cell-edit', (template, column) => {
 export class ColumnView {
 	constructor(model) {
 		this.model = model;
+
 		this.colspan = 1;
 		this.rowspan = 1;
+		this.index = -1;
```
titovmx, line 25: Should ```index``` property in ```ColumnModel``` be removed?
Also you should update d.ts files I guess

demo/pages/grouping-model/index.js
klumba12: this is just copy paste)

PR #423, opened 2017-08-29
Group model test

munkk:


src/core/group/group.model.spec.js
```diff
@@ -0,0 +1,13 @@
+import {Resource} from '../resource';
+import {Command} from '../command';
+import {GroupModel} from './group.model';
+
+describe('Group Model', () => {
+	it('constructor testing', () => {
+		let groupModel = new GroupModel();
+		expect(groupModel.resource).to.be.an.instanceOf(Resource);
```
klumba12, line 8: I think we can ommit tests for this file, it just check declaration

PR #427, opened 2017-08-30
Column factory test

munkk:


src/core/column/column.factory.spec.js
```diff
@@ -0,0 +1,20 @@
+import {columnFactory} from './column.factory';
```
klumba12, line 1: please add test to check body argument

src/core/column/column.factory.spec.js
```diff
@@ -0,0 +1,20 @@
...
+	let model = modelFactory();
+
+	it('should create a new instance of PasswordColumn', () => {
+		let foo = columnFactory(model);
+		let result = foo('password');
+		expect(result).to.be.an.instanceOf(PasswordColumn);
+	});
+
+	it('should create a new instance of CustomColumn if there is no such type in the ColumnType list', () => {
+		let foo = columnFactory(model);
```
klumba12: rename foo to createColumn

src/core/column/column.factory.spec.js
klumba12: let createColumn = columnFactory(model);
let result = createColumn('text', {path: 'foo', 'key': 'bar' etc.});

src/core/column/column.factory.spec.js
```diff
@@ -0,0 +1,28 @@
+import {columnFactory} from './column.factory';
+import {modelFactory} from '../test/model.factory';
+import {PasswordColumn} from '../column-type/password.column';
+import {ColumnView as CustomColumn} from '../column-type/column.model.view';
+
+describe('columnFactory', () => {
+	let model = modelFactory();
+	let body = {title: 'pass', pin: '1234'};
```
klumba12, line 8: assign pin: 'left' also check nested properties like editorOptions

PR #437, opened 2017-09-06
Unit factory test + dts

munkk:


src/core/dom/selector/unit.factory.d.ts
```diff
@@ -0,0 +1,9 @@
+import {Range} from '../../infrastructure/range';
+
+export declare class UnitFactory {
+	constructor(rowRange: Range, columnRange: Range);
+	rowRange: Range;
```
klumba12: rowRange and columnRange are private

klumba12 (approved):

PR #440, opened 2017-09-08
Selector.mark test + d.ts

munkk:


src/core/dom/selector/selector.mark.d.ts
```diff
@@ -0,0 +1,21 @@
+import {Model} from '../../infrastructure/model';
+import {Range} from '../../infrastructure/range';
+
+export interface ICell {
+	element: object;
+	columnRange: Range;
+	rowRange: Range;
+}
+
+export interface AddFactoryResult {
```
klumba12: please add `I` for all interfaces

src/core/dom/selector/selector.mark.d.ts
```diff
@@ -0,0 +1,21 @@
...
+	(pin: string | null): ICell[];
+}
+
+export declare class SelectorMark {
+	constructor(model: Model, markup: object, name: string);
+	model: Model;
```
klumba12: model markup name are private

klumba12: addFactory is also private

klumba12 (approved):

PR #447, opened 2017-10-08
Style + style.view fixes

klumba12:


src/core/infrastructure/model.d.ts
```diff
@@ -34,7 +34,7 @@ import {Event} from './event';
 export declare class Model {
 	constructor();
 
-	static register(name: string, model: Model): Model;
+	static register(name: string, model: {new(): {}}): Model;
```
etokareva, line 37: What is the purpose of new()? Don't like this name :-)

vkorolev: It means, that the second arg is a type that has parameterless constructor

src/core/style/style.view.js
```diff
@@ -110,7 +104,12 @@ export class StyleView extends View {
 					const cells = bodyRow.cells();
 					for (let j = 0, cellsLength = cells.length; j < cellsLength; j++) {
 						const cell = cells[j];
-						const column = columns[j];
+						const dataCell = cell.model();
+						if(!dataCell){
```
vkorolev: Space after `if` is missing:)

vkorolev (approved):

src/core/infrastructure/model.d.ts
etokareva: @vkorolev thanks

PR #457, opened 2017-11-02
Data manipulation style fix

vkorolev:


src/core/command/composite.d.ts
```diff
@@ -0,0 +1,7 @@
+import {Command} from './command';
```
klumba12: Please put composite to infrastructure

src/core/command/composite.js
```diff
@@ -0,0 +1,33 @@
+import {Command} from './command';
+import {noop} from '../utility';
+
+export class Composite {
+	static function (functions, reducer = noop, memo = null) {
```
klumba12: maybe func? function is so system :)

src/core/command/composite.js
```diff
@@ -0,0 +1,33 @@
...
+			const args = [].slice.apply(arguments);
+			for (let i = 0, length = functions.length; i < length; i++) {
+				const func = functions[i];
+				memo = reducer(memo, func.apply(null, args));
+			}
+			return memo;
+		};
+	}
+
+	static command() {
```
klumba12: use ...args

src/core/command/composite.js
klumba12: do we need reducer and memo

src/core/command/composite.js
```diff
@@ -0,0 +1,33 @@
+import {Command} from './command';
+import {noop} from '../utility';
+
+export class Composite {
+	static function (functions, reducer = noop, memo = null) {
+		return function () {
+			const args = [].slice.apply(arguments);
```
klumba12: ...args

vkorolev: yeah, good point. always forget about beautiful es6

src/core/command/composite.js
vkorolev: I think, in general case, it might be useful because potentially there can be a case when user needs to handle returned value somehow

klumba12 (approved):

PR #466, opened 2017-11-15
Validation init PR

etokareva:


src/plugins/validation/validator.js
```diff
@@ -0,0 +1,66 @@
+import PluginComponent from '../plugin.component';
+
+const LIVR = require('livr');
```
klumba12: use import

src/plugins/validation/validator.js
```diff
@@ -0,0 +1,66 @@
...
+
+	get resource() {
+		return this.model.validation().resource;
+	}
+
+	get rules() {
+		return this.model.validation().rules;
+	}
+
+	get validationArray() {
```
klumba12: please add validation.service and put method toLIVR there

src/core/infrastructure/model.d.ts
```diff
@@ -163,4 +163,8 @@ export declare class Model {
 	persistence(value?: object, tag?: object): FetchModel;
 
 	persistenceChanged: Event;
+
+	validation(value?: object, tag?: object): FetchModel;
```
titovmx: should return ```ValidationModel``` probably

src/plugins/validation/validator.js
```diff
@@ -0,0 +1,54 @@
...
+	onInit() {
+		const validationSettings = this.validationSettings;
+		if (validationSettings.canValidate) {
+			this.validator = new LIVR.Validator(validationSettings.rules);
+		}
+	}
+
+	get error() {
+		if (this.validator) {
+			const validated = {
```
klumba12: consider to rename it to target

src/plugins/validation/validator.js
```diff
@@ -0,0 +1,54 @@
...
+				[this.key]: this.value
+			};
+			const validData = this.validator.validate(validated);
```
klumba12: can we use only .getErrors()? there is no sence to call validate 2times?

src/plugins/validation/validator.js
```diff
@@ -0,0 +1,54 @@
...
+
+const Plugin = PluginComponent('validator');
+
+class Validator extends Plugin {
+	constructor() {
+		super(...arguments);
+	}
+
+	onInit() {
+		const validationSettings = this.validationSettings;
```
klumba12: i think it will be a little bit more readable to use service here explicitly and remove validationSettings

src/plugins/validation/validation.service.js
```diff
@@ -0,0 +1,18 @@
...
+				if (name !== 'key' && name !== 'for') {
+					validationRules.push({
+						[name]: rule[name]
+					});
+				}
+			}
+		}
+	});
+	return {
+		canValidate: validationRules.length > 0,
```
klumba12: rename it to hasRules

src/plugins/validation/validator.js
etokareva: I just thought that in the case of getter we could use these setting inside of other methods (possible future)

src/plugins/validation/validator.js
etokareva: Doesn't work like this

klumba12 (approved):

PR #473, opened 2017-11-17
Validation 0.1

etokareva:


src/plugins/validation/rule.js
```diff
@@ -7,17 +7,22 @@ class Rule extends Component {
 	}
 
 	onInit() {
+		const predefinedRules = ['length_between', 'length_equal', 'max_length', 'min_length', 'one_of', 'positive_integer', 'required'];
```
klumba12: is it possible go through $attrs and exclude "key" and "for"? or even better to have on the top of file
```javascript
ruleBindings = [
{length-beetwen: '<? ...'}
]
```
and use it in onInit and in bindings?

etokareva: Actually, I had the same thought before 👍 But forgot about it.
I'll do small research 😄

vkorolev: What is going to happen to typos/errors in this case?

klumba12 (approved):

PR #477, opened 2017-11-20
use focusin for active state indication

klumba12:


src/view/components/grid/grid.js
```diff
@@ -138,8 +143,15 @@ export class Grid extends RootComponent {
 		return this.model.visibility();
 	}
 
-	get isActive() {
-		return this.table.view.isFocused();
+	invalidateIsActive() {
```
vkorolev: Maybe just `invalidateActive`?

klumba12: maybe even better name?

klumba12: What is better? onFocusChanged or invalidateActive? onFocusin, captureFocus 
any suggestions?

vkorolev (approved):

PR #480, opened 2017-11-20
Prevent scrolling when space or shift+space 

klumba12:


vkorolev (approved):

src/core/navigation/navigation.js
```diff
@@ -114,102 +115,214 @@ export class Navigation {
...
 				shortcut: shortcut.up,
-				canExecute: () => canNavigate() && this.prevRow >= 0,
-				execute: () => this.goTo(this.prevRow, this.currentColumn)
+				canExecute: () => {
+					if (canNavigate()) {
+						const newRow = this.prevRow;
+						return newRow >= 0 && go.canExecute(this.context('up', {newRow}));
+					}
+
+					return false;
```
etokareva, line 175: do we need to return false explicitly?

PR #481, opened 2017-11-21
Editor position regarding window rect

klumba12:


src/core/navigation/navigation.js
```diff
@@ -114,102 +115,214 @@ export class Navigation {
...
 				shortcut: shortcut.down,
-				canExecute: () => canNavigate() && this.nextRow >= 0,
-				execute: () => this.goTo(this.nextRow, this.currentColumn)
+				canExecute: () => {
+					if (canNavigate()) {
+						const newRow = this.nextRow;
+						return newRow >= 0 && go.canExecute(this.context('down', {newRow}));
+					}
+
+					return false;
```
etokareva, line 159: Do we need to return false explicitly here? Without this line in case of !canNavigate() function canExecute will return 'undefined' by default

src/view/directives/position.js
```diff
@@ -71,8 +72,12 @@ class Position extends Directive(POSITION_NAME, {root: `^?${GRID_NAME}`}) {
 		const intersection = max(intersections, i => i.area);
 		const rect = intersection.b;
 
-		source.style.left = tr.left + (rect.left - tr.left) + 'px';
-		source.style.top = tr.top + (rect.top - tr.top) + 'px';
+		const left = tr.left + (rect.left - tr.left);
```
titovmx: Hmm...I tried to open brackets and looks that something is redundant :) just ```const left = rect.left```

src/core/navigation/navigation.js
klumba12: yes because command manager checks explictly for false in some cases

src/view/directives/position.js
klumba12: fuck yeah)

klumba12: actually this will be not the last version of layout algo)

src/core/navigation/navigation.js
```diff
@@ -114,102 +115,214 @@ export class Navigation {
...
+				},
+				execute: () => {
+					const newRow = this.nextRow;
+					const newColumn = this.currentColumn;
+					return go.execute(this.context('down', {newRow, newColumn})) && this.goTo(newRow, newColumn);
```
titovmx, line 164: it's minor, but we still have incorrect interface for ```execute``` method in d.ts file which returns void.

titovmx (approved):

src/view/directives/position.js
etokareva: I suppose it should be updated both here and in #482

PR #488, opened 2017-11-21
Put validation under commit.canExecute

etokareva:


src/plugins/validation/validator.js
```diff
@@ -1,5 +1,5 @@
 import PluginComponent from '../plugin.component';
-import * as validationService from './validation.service';
+import * as validationService from '@grid/core/validation/validation.service';
```
vkorolev, line 2: is it more readable?:)

src/core/edit/edit.cell.view.js
```diff
@@ -105,15 +106,19 @@ export class EditCellView extends View {
 				priority: 1,
 				source: 'edit.cell.view',
 				shortcut: this.shortcutFactory('commit'),
-				// TODO: add validation support
 				canExecute: cell => {
 					cell = cell || this.editor.cell;
+					const context = this.contextFactory(cell);
+					const key = context.column.key;
+					const validator = validationService.createValidator(model.validation().rules, key);
+					const valid = validator.validate({[key]: this.value});
```
klumba12: this should be the in the last && so there is a chance to not call it

src/plugins/validation/validator.js
etokareva: No)) But I moved validation.service to core. So it is more readable than '../../core/validation/validation.service'

vkorolev: Oh, i see, sorry

vkorolev (approved):

PR #491, opened 2017-11-22
relayout editor on window resize

klumba12:


src/view/directives/position.js
```diff
@@ -10,6 +12,15 @@ class Position extends Directive(POSITION_NAME, {root: `^?${GRID_NAME}`}) {
 		this.$attrs = $attrs;
 		this.$timeout = $timeout;
 		this.$window = $window;
+		
+		const listener = new EventListener(this.$window, new EventManager(this));
+		const job = jobLine(400);
+		
+		this.using(listener.on('resize', () => {
+			this.onInit();
```
vkorolev: It's not a big deal, but `onInit` should be called only once. it would be better to extract `onInit` to another method and call it instead.

vkorolev (approved):

PR #494, opened 2017-11-22
Fix karma and chrome launcher versions.

vkorolev:


package.json
```diff
@@ -60,10 +60,10 @@
     "json-loader": "^0.5.4",
     "jspdf": "^1.3.3",
     "jspdf-autotable": "^2.3.2",
-    "karma": "^1.2.0",
+    "karma": "^1.7.0",
```
etokareva, line 63: I'm not sure. Should we update karma-webpack also?

etokareva: No, it's OK

etokareva (approved):

PR #495, opened 2017-11-23
Validation defaults + fix context

etokareva:


src/core/edit/edit.cell.view.js
```diff
@@ -109,15 +109,19 @@ export class EditCellView extends View {
 				canExecute: cell => {
 					cell = cell || this.editor.cell;
 					const context = this.contextFactory(cell);
-					const key = context.column.key;
-					const validator = validationService.createValidator(model.validation().rules, key);
+					let isValid = true;
+					if (context){
```
vkorolev: Space is missing  `) {`

src/core/edit/edit.cell.view.js
```diff
@@ -260,6 +264,9 @@ export class EditCellView extends View {
 	}
 
 	contextFactory(cell, value, label, tag) {
+		if (!cell) {
```
klumba12: this is not right context should always return context not boolean, if cell is undefined createContext shouln't be called

src/core/edit/edit.cell.view.js
etokareva: :-)

src/core/edit/edit.cell.view.js
```diff
@@ -108,16 +108,18 @@ export class EditCellView extends View {
 				shortcut: this.shortcutFactory('commit'),
 				canExecute: cell => {
 					cell = cell || this.editor.cell;
-					const context = this.contextFactory(cell);
-					const key = context.column.key;
-					const validator = validationService.createValidator(model.validation().rules, key);
-					return cell
+					const isEditableCell = cell
```
klumba12: rename to canEdit

klumba12 (approved):

PR #496, opened 2017-11-23
Data manipulation plugin refactoring

klumba12:


plugin/data-manipulation/data.manipulation.view.js
```diff
@@ -0,0 +1,234 @@
...
+					},
+					canExecute: e => {
+						const rowId = this.rowId(e.row);
+						return this.changes.deleted.has(rowId) || this.changes.edited.has(rowId);
+					}
+				}),
+				'Revert Row',
+				'restore'
+			),
+			// new Action(
```
vkorolev, line 142: Is this something we need?

klumba12: Yes we need to have a service that will allow to open edit form from the user code

plugin/data-manipulation/data.manipulation.view.js
```diff
@@ -0,0 +1,234 @@
...
+import {set as setLabel} from '@grid/core/services/label';
+import {PluginView} from '../plugin.view';
+
+export class DataManipulationView extends PluginView {
+	constructor(model) {
+		super();
+
+		this.model = model;
+
+		this.commitCommand = new Command({
```
etokareva, line 16: The main part of this code is from src/plugins/data-manipulation/data.manipulation.js file. Maybe it would be better to move src/plugins/data-manipulation/data.manipulation.js to this directory and name it data.manipulation.view.js to keep history of changes and in src/plugins/data-manipulation/ create new data.manipulation.js file. Does it make sence?

titovmx (approved):

plugin/data-manipulation/data.manipulation.view.js
klumba12: Yes but a standard naming for such of services that we're using ends up with .view

PR #498, opened 2017-11-23
output list-of items errors

etokareva:


demo/pages/validation/index.html
```diff
@@ -34,5 +36,13 @@
 		             key="contact.address.state"
 		             length-equal="2">
 		</q-grid:rule>
+		<q-grid:rule for="column"
```
klumba12: do we need to changed for="column" to for="cell"?

src/plugins/validation/validator.js
```diff
@@ -21,7 +23,18 @@ class Validator extends Plugin {
 			};
 			const validData = this.validator.validate(target);
```
klumba12: isValid

src/plugins/validation/validator.js
```diff
@@ -21,7 +23,18 @@ class Validator extends Plugin {
...
 			if (!validData) {
-				return this.validator.getErrors()[this.key];
+				const errors = this.validator.getErrors()[this.key];
+				if (isString(errors)) {
+					return errors;
+				} else if (isArray(errors) && isArray(this.value)) {
+					const result = errors.reduce((prev, curr, i) => {
```
klumba12: prev - memo
curr - error

src/plugins/validation/validator.js
```diff
@@ -21,7 +23,18 @@ class Validator extends Plugin {
...
+						if (curr !== null) {
+							prev.push(`${this.value[i]}: ${curr}`);
+						}
+						return prev;
+					}, []);
+					return result.join('\n');
```
klumba12: can we always return array of strings? so UI template can decide how to show this?

demo/pages/validation/index.html
etokareva: I can change, OK

src/plugins/validation/validator.js
etokareva: Yes, we can. it would be better

src/plugins/validation/validator.js
```diff
@@ -1,6 +1,8 @@
 import PluginComponent from '../plugin.component';
 import * as validationService from '@grid/core/validation/validation.service';
+import {isString} from '@grid/core/utility';
 
+const isArray = angular.isArray;
```
titovmx: Why don't you use ```isArray``` from utility?

etokareva: This is a good question about isArray. Firtly, I used Array.isArray ES6 method, but our interpreter cursed on me and showed me that I should use angular.isArray instead. So, I used angular.isArray and it is OK for now. But maybe in the future this code will be moved to the core. So this is reasonable to use utility isArray method there.

klumba12: @grid/core/utility contains isArray

etokareva: I know

klumba12 (approved):

PR #499, opened 2017-11-24
refactor persistence plugin

titovmx:


plugin/persistence/persistence.view.js
```diff
@@ -0,0 +1,191 @@
...
+		this.state = {
+			editItem: null,
+			oldValue: null
+		};
+		this.model = model;
+		this.storageKey = `q-grid:${model.grid().id}:${model.persistence().id}:persistence-list`;
+		this.persistenceService = new PersistenceService(model);
+		this.submitEvent = new Event();
+		this.cancelEvent = new Event();
+		this.closeEvent = new Event();
```
klumba12: lets use ony cancelEvent, to close or not close decides plugin itself

src/plugins/persistence/persistence.js
```diff
@@ -48,14 +37,13 @@ class Persistence extends Plugin {
 
 						const config = {
 							attachTo: angular.element(this.$document[0].body), // eslint-disable-line no-undef
-							controller: PersistencePanelController,
+							controller: () => {},
```
klumba12: can we just use persistenceView in controller and inside of controller make $scope.persistence = new PersView()?
so no need to write $persistence.view?

plugin/persistence/persistence.view.js
```diff
@@ -0,0 +1,191 @@
...
+	isActive(item) {
+		return JSON.stringify(item.model) === JSON.stringify(this.persistenceService.save()); // eslint-disable-line angular/json-functions
+	}
+
+	stringify(item) {
+		const model = item.model;
+		const targets = [];
+		const settings = this.model.persistence().settings;
+
+		for (let key in settings) {
```
etokareva, line 170: do we want to check all settings properties there or own?

titovmx: all iterable properties exactly

klumba12 (approved):

PR #505, opened 2017-11-27
Column filter view

titovmx:


plugin/column-filter/column.filter.view.js
```diff
@@ -0,0 +1,114 @@
...
+export class ColumnFilterView extends PluginView {
+	constructor(model, context) {
+		super(...arguments);
+
+		this.model = model;
+		this.key = context.key;
+
+		this.cancelEvent = new Event();
+		this.submitEvent = new Event();
+		this.resetItemsEvent = new Event();
```
klumba12: rename it to reset and don't call in constructor

klumba12 (approved):

PR #511, opened 2017-11-29
put action to core plugins

titovmx:


src/plugins/action/action.js
```diff
@@ -12,42 +12,37 @@ TemplatePath
...
 	}
 
 	onInit() {
-		const model = this.model;
-		const action = new ActionItem(this.command, this.title, this.icon);
-		action.id = this.id;
-
-		const actions = Array.from(model.action().items);
-		actions.push(action);
+		const context = {
```
klumba12, line 22: u can make like `ActionView(this.model, this)` and it will persist bindings

klumba12 (approved):

src/plugins/action/action.js
titovmx: I'm not sure that it's good from typification perspective

PR #512, opened 2017-11-29
Mouse wheel fixes & Scroll perf improvements + npm i vscroll (new version)

klumba12:


src/core/infrastructure/event.listener.js
```diff
@@ -5,12 +5,12 @@ export class EventListener {
 		this.handlers = {};
 	}
 
-	on(name, f) {
+	on(name, f, settings = null) {
```
etokareva: Can we set settings=false?

titovmx (approved):

src/core/infrastructure/event.listener.js
titovmx: or/and change variable name, as I understand it's just boolean ```useCapture``` argument of ```addEventListener```. Does it make sense?

etokareva: I mean default parameter applies when value is not passed, we set it to ``null`` and after if null we use false. Why?  We can just set default to false. And ``this.element.addEventListener(name, handler, settings);``

@titov-max here settings or options is correct because of new feature - passive event listeners and EventListenerOptions object, and going with it - [Event listener options](https://dom.spec.whatwg.org/#dom-addeventlisteneroptions-passive)

vkorolev: @titov-max that's a good point. Old standards defined third parameter as boolean `useCapture`. But now it can be [a hash](https://dom.spec.whatwg.org/#interface-eventtarget) . I think this should be noticed, because it may not work with  <51 Chrome versions

titovmx: Oh, right, I see. Thank you, folks. A hash confuses a bit really.

titovmx: @vkorolev, @titov-max Without polyfill this object will be coerced to true so event listener will be executed in capturing phase. It's not big deal (or am I wrong?), because it will be triggered only in old browsers. But we can add add two-rows-length polyfill with check of Object.defineProperty for the case of supporting object props and check of type if it's important

PR #514, opened 2017-11-30
put cell editor to core

titovmx:


klumba12 (approved):

plugin/cell-editor/cell.editor.view.js
```diff
@@ -0,0 +1,15 @@
+import {PluginView} from '../plugin.view';
+import {Event} from '@grid/core/infrastructure';
+
+export class CellEditorView extends PluginView {
+	constructor(model) {
+		super(...arguments);
```
etokareva, line 6: If you not planning to change number of constructor arguments, then we could use ``model`` instead

src/plugins/cell-editor/cell.editor.js
```diff
@@ -1,28 +1,29 @@
 import Component from '../../view/components/component';
 import {VIEW_CORE_NAME} from '@grid/view/definition';
+import {CellEditorView} from '@grid/plugin/cell-editor/cell.editor.view';
 
 class CellEditor extends Component {
-	constructor() {
+	constructor($scope) {
 		super();
+
+		this.$scope = $scope;
```
etokareva, line 9: can we use ``super($scope)`` to set this.$scope to $scope or it is TS feature?

etokareva: I've just checked. We can.

PR #516, opened 2017-11-30
export refactoring

etokareva:


plugin/export/export.view.js
```diff
@@ -0,0 +1,91 @@
...
+
+	get rows() {
+		return this.model.data().rows;
+	}
+
+	get columns() {
+		return this.model.data().columns;
+	}
+
+	get resource() {
```
klumba12: this should be outside of view

klumba12 (approved):

PR #517, opened 2017-11-30
put edit form to core

titovmx:


plugin/edit-form/edit.form.panel.view.js
```diff
@@ -0,0 +1,65 @@
...
+import {isUndefined} from '@grid/core/utility';
+import {Command} from '@grid/core/command';
+import {RowEditor} from '@grid/core/edit/edit.row.editor';
+import {Event} from '@grid/core/infrastructure';
+
+export class EditFormPanelView extends PluginView {
+	constructor(model, context) {
+		super(model);
+		
+		this.model = model;
```
etokareva: I don't see the reason to do this explicitly

etokareva (approved):

plugin/edit-form/edit.form.panel.view.js
titovmx: good point

klumba12 (approved):

PR #518, opened 2017-11-30
Scene stop scheduling improvement

klumba12:


src/core/infrastructure/log.d.ts
```diff
@@ -1,6 +1,9 @@
 export declare class Log {
 	constructor();
+
```
etokareva, line 3: Can we remove this and also add eslint rule no-useless-constructor?

klumba12: no to write constructor is a good style even empty, at least for me

etokareva: @klumba12 ok 😸

etokareva (approved):

PR #524, opened 2017-12-04
Put core outside of src + make *.ctrl services

klumba12:


src/view/components/body/td.core.js
```diff
@@ -2,8 +2,9 @@ import Directive from '@grid/view/directives/directive';
 import cellBuilder from '../cell/cell.build';
 import {AppError} from '@grid/core/infrastructure';
 import {VIEW_CORE_NAME, TD_CORE_NAME, GRID_NAME} from '@grid/view/definition';
-import {GRID_PREFIX} from '@grid/core/definition';
-import * as css from '@grid/core/services/css';
+import {TdCtrl} from '@grid/core/cell/td.ctrl';
+ 
+const classify = TdCtrl.classify;
```
vkorolev, line 7: Somewhere `TdCtrl` methods are called through class reference e.g `TdCtrl.classify(...)`, and somewhere through the variable: `const classify = TdCtrl.classify` .
I think it should be unified

core/style/style.view.d.ts
```diff
@@ -20,5 +20,6 @@ export declare class StyleView extends View {
 	valueFactory: IValueFactory;
 	active: IActive;
 	monitor: IMonitor;
-	invalidate(): void;
+	invalidate(any, any): void;
```
titovmx: argument names?

bower.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "ng-qgrid",
-	"version": "1.0.63",
+	"version": "1.3.0",
```
titovmx: Are you going to release new version?

src/view/components/body/td.core.js
klumba12: yes but here i need performance we can think about this in the future

vkorolev (approved):

titovmx (approved):

PR #530, opened 2017-12-07
Qgrid 43/expression filter plugin

vkorolev:


src/plugins/expression-builder/where-schema.js
```diff
@@ -0,0 +1,298 @@
...
+
+								return (field && field.type) || 'text';
+							},
+							change: function (node, line) {
+								if (node.attr('placeholder')) {
+									node.addAfter(node.clone());
+									node.attr('placeholder', false);
+								}
+
+								var field = this.value,
```
titovmx: Can ```const``` be used here?
Also formatting doesn't look pretty

src/view/directives/autosize.js
```diff
@@ -0,0 +1,76 @@
...
+	onInit() {
+		let $element = this.$element,
+			document = this.$document[0],
+			body = document.body,
+			selector = this.$attrs.autosize,
+			focusElement = null,
+			actualWidth = 0,
+			actualText = '';
+
+		function calculateWidth(text) {
```
titovmx: Can we put this method to core view for autosize to reuse it in the future for other frameworks? Like position.view for position directive.

src/view/directives/autosize.js
```diff
@@ -0,0 +1,76 @@
...
+				body.append(container);
+				// TODO: cache widths be text ?
+				width = test.offsetWidth;
+				container.remove();
+			}
+
+			return width;
+		}
+
+		this.width = function (text) {
```
titovmx: Looks like it can be a method of ```Autosize``` and many variables declared above are using just in the ```this.width``` function. So maybe we can declare them here. 
And can you prompt me where is using of ```this.width```?

src/plugins/expression-builder/where-schema.js
vkorolev: Yes, I'll use const. 
what's wrong with formatting?

src/view/directives/autosize.js
vkorolev: Well, basically we can, but right now it's used only in expression-builder, which is tightly coupled to angularjs. I think we can do it in case it is used within the grid.

src/view/directives/autosize.js
vkorolev: it's used in autocomplete template: `ng-style="{'width' :  $autosize.width(expression.value) + 'px'}"`

src/plugins/expression-builder/where-schema.js
titovmx: Cause of you write variables divided by comma, next row variable shifted to right by double tab. Cause of we're using tabs in the project, in github review it's located far from first variable row declaration. Really it's not a big problem, more my personal hating of tab formatting issues :)

src/view/directives/autosize.js
titovmx: Yeah, I got. Probably now this refactoring is redundant.

src/plugins/expression-builder/where-schema.js
vkorolev: Yeah, agree, there's some pain about reviewing such structures.

titovmx (approved):

package.json
```diff
@@ -22,9 +22,10 @@
   "homepage": "https://github.com/qgrid/ng#readme",
   "dependencies": {
     "angular": "^1.5.8",
-    "angular-messages": "^1.5.8",
     "angular-material": "^1.1.4",
-    "angular-sanitize": "^1.5.8"
+    "angular-messages": "^1.5.8",
+    "angular-sanitize": "^1.5.8",
+    "ng-expression-builder": "^1.0.0"
```
klumba12, line 28: can u put it to dev dependencies? it will be a core plugin that will be included in the source code itself

src/plugins/expression-builder/operations.js
```diff
@@ -0,0 +1,78 @@
+const swap = function (obj) {
```
klumba12, line 1: can u put it under root plugin folder? not a ng plugin

klumba12 (approved):

PR #533, opened 2017-12-17
Column sort component + added dts files

klumba12:


src/plugins/column-sort/column.sort.js
```diff
@@ -0,0 +1,46 @@
...
+
+ColumnSort.$inject = ['$element'];
+
+export default {
+	transclude: true,
+	templateUrl: 'qgrid.plugin.column-sort.tpl.html',
+	controller: ColumnSort,
+	controllerAs: '$columnSortPlugin',
+	bindings: {
+		'column': '<column'
```
etokareva: could we write this shorter?

src/plugins/column-sort/column.sort.js
```diff
@@ -0,0 +1,46 @@
...
+import {VIEW_CORE_NAME} from '@grid/view/definition';
+
+import {ColumnSortView} from '@grid/plugin/column-sort/column.sort.view';
+
+class ColumnSort extends Component {
+	constructor($element) {
+		super();
+
+		this.element = $element[0];
+		this.descIcon = null;
```
titovmx: Looks like ```this.descIcon``` and ```this.ascIcon``` isn't used

titovmx (approved):

etokareva (approved):

PR #547, opened 2018-02-07
Column view width

klumba12:


src/view/components/cell/cell.handler.js
```diff
@@ -14,26 +14,49 @@ class CellHandler extends Component {
...
+
+		// When navigate first or when animation wasn't applied we need to omit
+		// next navigation event to make handler to correct position.
+		let isValid = false;
+		model.navigationChanged.watch(e => {
+			if (e.hasChanges('cell')) {
+				const domCell = table.body.cell(e.state.rowIndex, e.state.columnIndex);
+				const cell = e.state.cell;
+				if (cell) {
+					const oldColumn = (e.changes.cell.oldValue || {}).column || {};
```
etokareva: Confused a bit with these lines

klumba12: oldValue can be null so to get something not nullable to oldColumn we write like that

klumba12: any other ideas?

etokareva: @klumba12 Maybe 
`` e.changes.cell.oldValue ? e.changes.cell.oldValue.column : {} ``
or strict null check in the first place

titovmx (approved):

